PR #2, opened 2020-03-04
Laboratorium 1 - Zadanie 3.2

raionsakana:
Wykonano:
- poprawienie nazwy klasy MyAdapter,
- poprawienie rysowania,
- rysowanie wewnątrz okna aplikacji

### DIFF
--- a/src/main/java/edu/kis/powp/jobs2d/drivers/adapter/DrawPanelAdapter.java
+++ b/src/main/java/edu/kis/powp/jobs2d/drivers/adapter/DrawPanelAdapter.java
@@ -9,10 +9,10 @@ import edu.kis.powp.jobs2d.features.DrawerFeature;
 /**
  * driver adapter to drawer with several bugs.
  */
-public class MyAdapter extends DrawPanelController implements Job2dDriver {
+public class DrawPanelAdapter extends DrawPanelController implements Job2dDriver {
 	private int startX = 0, startY = 0;
 
-	public MyAdapter() {
+	public DrawPanelAdapter() {
 		super();
 	}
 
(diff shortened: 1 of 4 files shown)


src/main/java/edu/kis/powp/jobs2d/drivers/adapter/DrawPanelAdapter.java
```diff
@@ -4,14 +4,15 @@
 import edu.kis.legacy.drawer.shape.ILine;
 import edu.kis.legacy.drawer.shape.LineFactory;
 import edu.kis.powp.jobs2d.Job2dDriver;
+import edu.kis.powp.jobs2d.features.DrawerFeature;
 
 /**
  * driver adapter to drawer with several bugs.
  */
-public class MyAdapter extends DrawPanelController implements Job2dDriver {
+public class DrawPanelAdapter extends DrawPanelController implements Job2dDriver {
```
kowallus, line 12: The name is much better. Only an aspect concerning driver API is missing.

src/main/java/edu/kis/powp/jobs2d/drivers/adapter/DrawPanelAdapter.java
```diff
@@ -27,7 +28,8 @@ public void operateTo(int x, int y) {
 		line.setStartCoordinates(this.startX, this.startY);
 		line.setEndCoordinates(x, y);
 
-		drawLine(line);
+		setPosition(x, y);
+		DrawerFeature.getDrawerController().drawLine(line);
```
kowallus, line 32: This is unnecessary direct coupling with DrawerFeature class!

kowallus (commented):

The name of a commit: "Figure Joe 2 added to app" does not correspond to its contents.

PR #4, opened 2020-03-04
Lab1 - 215693

maybegamer:
Zakonczone na podpunkcie 3.2.1

### DIFF
--- a/src/main/java/edu/kis/powp/jobs2d/drivers/adapter/DrawPanelAdapter.java
+++ b/src/main/java/edu/kis/powp/jobs2d/drivers/adapter/DrawPanelAdapter.java
@@ -10,10 +10,10 @@ import edu.kis.powp.jobs2d.features.DriverFeature;
 /**
  * driver adapter to drawer with several bugs.
  */
-public class MyAdapter extends DrawPanelController implements Job2dDriver {
+public class DrawPanelAdapter extends DrawPanelController implements Job2dDriver {
 	private int startX = 0, startY = 0;
 
-	public MyAdapter() {
+	public DrawPanelAdapter() {
 		super();
 	}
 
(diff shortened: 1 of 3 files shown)


src/main/java/edu/kis/powp/jobs2d/drivers/adapter/DrawPanelAdapter.java
```diff
@@ -4,14 +4,16 @@
...
 import edu.kis.legacy.drawer.shape.LineFactory;
 import edu.kis.powp.jobs2d.Job2dDriver;
+import edu.kis.powp.jobs2d.features.DrawerFeature;
+import edu.kis.powp.jobs2d.features.DriverFeature;
 
 /**
  * driver adapter to drawer with several bugs.
  */
-public class MyAdapter extends DrawPanelController implements Job2dDriver {
+public class DrawPanelAdapter extends DrawPanelController implements Job2dDriver {
```
kowallus, line 13: The name is much better. Only an aspect concerning driver API is missing.

src/main/java/edu/kis/powp/jobs2d/drivers/adapter/DrawPanelAdapter.java
```diff
@@ -27,7 +29,9 @@ public void operateTo(int x, int y) {
 		line.setStartCoordinates(this.startX, this.startY);
 		line.setEndCoordinates(x, y);
 
-		drawLine(line);
+		setPosition(x, y);
+
+		DrawerFeature.getDrawerController().drawLine(line);
```
kowallus, line 34: This is unnecessary direct coupling with DrawerFeature class!

src/test/java/edu/kis/powp/jobs2d/TestJobs2dPatterns.java
```diff
@@ -83,7 +83,7 @@ public static void main(String[] args) {
 			public void run() {
 				Application app = new Application("2d jobs Visio");
 				DrawerFeature.setupDrawerPlugin(app);
-				setupDefaultDrawerVisibilityManagement(app);
+				//setupDefaultDrawerVisibilityManagement(app);
```
kowallus, line 86: In revision repo it is safe to delete code instead of using a comment.

kowallus (commented):

PR #5, opened 2020-03-04
instrukcja adapter 3.2.2 skończone

Grzegorian:


src/main/java/edu/kis/powp/jobs2d/drivers/adapter/DriverAdapter.java
```diff
@@ -4,34 +4,39 @@
 import edu.kis.legacy.drawer.shape.ILine;
 import edu.kis.legacy.drawer.shape.LineFactory;
 import edu.kis.powp.jobs2d.Job2dDriver;
+import edu.kis.powp.jobs2d.features.DrawerFeature;
 
 /**
  * driver adapter to drawer with several bugs.
  */
-public class MyAdapter extends DrawPanelController implements Job2dDriver {
+public class DriverAdapter extends DrawPanelController implements Job2dDriver {
```
kowallus, line 12: The name is better but it does not indicate the purpose of the adapter or what was actually adapted.

src/main/java/edu/kis/powp/jobs2d/drivers/adapter/DriverAdapter.java
```diff
@@ -4,34 +4,39 @@
...
+	
 	@Override
 	public void operateTo(int x, int y) {
 		ILine line = LineFactory.getBasicLine();
 		line.setStartCoordinates(this.startX, this.startY);
 		line.setEndCoordinates(x, y);
-
-		drawLine(line);
+		startX = x;
+		startY = y;
```
kowallus, line 34: Minor code duplication

src/main/java/edu/kis/powp/jobs2d/drivers/adapter/DriverAdapter.java
```diff
@@ -4,34 +4,39 @@
...
+public class DriverAdapter extends DrawPanelController implements Job2dDriver {
 	private int startX = 0, startY = 0;
-
-	public MyAdapter() {
+	private DrawPanelController drawerFeature;
+	
+	
+	public DriverAdapter() {
 		super();
+		drawerFeature = DrawerFeature.getDrawerController();
```
kowallus, line 19: This is unnecessary direct coupling with DrawerFeature class!

kowallus (commented):

PR #8, opened 2020-03-11
Lab 1 zad 3.2 obecność

cyngiel:


.idea/codeStyles/codeStyleConfig.xml
```diff
@@ -0,0 +1,5 @@
+<component name="ProjectCodeStyleConfiguration">
```

src/main/java/edu/kis/powp/jobs2d/drivers/adapter/MyAdapter.java
```diff
@@ -27,7 +28,9 @@ public void operateTo(int x, int y) {
 		line.setStartCoordinates(this.startX, this.startY);
 		line.setEndCoordinates(x, y);
 
-		drawLine(line);
+		DrawerFeature.getDrawerController().drawLine(line);
+		this.startX = x;
+		this.startY = y;
```
kowallus, line 33: Minor code duplication

src/main/java/edu/kis/powp/jobs2d/drivers/adapter/MyAdapter.java
```diff
@@ -27,7 +28,9 @@ public void operateTo(int x, int y) {
 		line.setStartCoordinates(this.startX, this.startY);
 		line.setEndCoordinates(x, y);
 
-		drawLine(line);
+		DrawerFeature.getDrawerController().drawLine(line);
```
kowallus, line 31: This is unnecessary direct coupling with DrawerFeature class!

src/test/java/edu/kis/powp/jobs2d/TestJobs2dPatterns.java
```diff
@@ -83,7 +83,7 @@ public static void main(String[] args) {
 			public void run() {
 				Application app = new Application("2d jobs Visio");
 				DrawerFeature.setupDrawerPlugin(app);
-				setupDefaultDrawerVisibilityManagement(app);
+				//setupDefaultDrawerVisibilityManagement(app);
```
kowallus, line 86: In revision repo it is safe to delete code instead of using a comment.

kowallus (commented):

PR #10, opened 2020-03-14
Refactor final 3.2

GGuinea:


src/test/java/edu/kis/powp/jobs2d/events/SelectTestFigureOptionListener.java
```diff
@@ -3,19 +3,29 @@
...
 
-	public SelectTestFigureOptionListener(DriverManager driverManager) {
+	private java.lang.reflect.Method method = null;
+
+	private FiguresJoe obj = new FiguresJoe();
+
+	public SelectTestFigureOptionListener(DriverManager driverManager, String figureScript) {
 		this.driverManager = driverManager;
+		try {
+		  method = obj.getClass().getMethod(figureScript, (Job2dDriver.class));
```
kowallus: Reflection is not object-oriented and type-safe mechanisms. Adding a new listener class would be consistent with an approach used in the project.

GGuinea: I've chosen enum instead of listener. I don't now which option is better.

src/main/java/edu/kis/powp/jobs2d/drivers/adapter/DrawController.java
```diff
@@ -4,6 +4,7 @@
 import edu.kis.legacy.drawer.shape.ILine;
 import edu.kis.legacy.drawer.shape.LineFactory;
 import edu.kis.powp.jobs2d.Job2dDriver;
+import edu.kis.powp.jobs2d.features.DrawerFeature;
```
kowallus: Is this line used?

src/test/java/edu/kis/powp/jobs2d/events/SelectTestFigureOptionListener.java
kowallus: Much better now :)

src/main/java/edu/kis/powp/jobs2d/drivers/adapter/DrawController.java
GGuinea: My bad, removed.

PR #11, opened 2020-03-17
Zadania z adaptera

BartoszKowalczyk98:
indeks: 215772

### DIFF
--- a/src/main/java/edu/kis/powp/jobs2d/drivers/adapter/DrawControllerAdapter.java
+++ b/src/main/java/edu/kis/powp/jobs2d/drivers/adapter/DrawControllerAdapter.java
@@ -5,13 +5,15 @@ import edu.kis.legacy.drawer.shape.ILine;
 import edu.kis.legacy.drawer.shape.LineFactory;
 import edu.kis.powp.jobs2d.Job2dDriver;
 
+import static edu.kis.powp.jobs2d.features.DrawerFeature.getDrawerController;
+
 /**
  * driver adapter to drawer with several bugs.
  */
-public class MyAdapter extends DrawPanelController implements Job2dDriver {
+public class DrawControllerAdapter extends DrawPanelController implements Job2dDriver {
 	private int startX = 0, startY = 0;
 
-	public MyAdapter() {
+	public DrawControllerAdapter() {
 		super();
 	}
 
@@ -27,7 +29,8 @@ public class MyAdapter extends DrawPanelController implements Job2dDriver {
 		line.setStartCoordinates(this.startX, this.startY);
 		line.setEndCoordinates(x, y);
 
-		drawLine(line);
+		getDrawerController().drawLine(line);
+		setPosition(x,y);
 	}
 
 	@Override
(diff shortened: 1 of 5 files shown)


src/main/java/edu/kis/powp/jobs2d/drivers/adapter/DrawControllerAdapter.java
```diff
@@ -5,13 +5,15 @@
 import edu.kis.legacy.drawer.shape.LineFactory;
 import edu.kis.powp.jobs2d.Job2dDriver;
 
+import static edu.kis.powp.jobs2d.features.DrawerFeature.getDrawerController;
+
 /**
  * driver adapter to drawer with several bugs.
  */
-public class MyAdapter extends DrawPanelController implements Job2dDriver {
+public class DrawControllerAdapter extends DrawPanelController implements Job2dDriver {
```
kowallus: The name is better but it does not indicate the purpose of the adapter or what was actually adapted.

Inheritance is not used (no super drawLine calls).

src/main/java/edu/kis/powp/jobs2d/drivers/adapter/DrawControllerAdapter.java
```diff
@@ -5,13 +5,15 @@
 import edu.kis.legacy.drawer.shape.LineFactory;
 import edu.kis.powp.jobs2d.Job2dDriver;
 
+import static edu.kis.powp.jobs2d.features.DrawerFeature.getDrawerController;
```
kowallus: Direct coupling with DrawerFeature class is unnecessary.

src/main/java/edu/kis/powp/jobs2d/drivers/adapter/LineDrawerAdapter.java
```diff
@@ -0,0 +1,35 @@
+package edu.kis.powp.jobs2d.drivers.adapter;
+
+import edu.kis.legacy.drawer.panel.DrawPanelController;
+import edu.kis.legacy.drawer.shape.ILine;
+import edu.kis.legacy.drawer.shape.LineFactory;
+import edu.kis.powp.jobs2d.Job2dDriver;
+
+import static edu.kis.powp.jobs2d.features.DrawerFeature.getDrawerController;
+
+public class LineDrawerAdapter extends DrawPanelController implements Job2dDriver{
```
kowallus: Similar issues.

src/test/java/edu/kis/powp/jobs2d/events/SelectTestFigureOptionListener.java
```diff
@@ -16,6 +16,11 @@ public SelectTestFigureOptionListener(DriverManager driverManager) {
 
 	@Override
 	public void actionPerformed(ActionEvent e) {
-		FiguresJoe.figureScript1(driverManager.getCurrentDriver());
+
+		if(e.paramString().contains("Figure Joe 1"))
```

src/main/java/edu/kis/powp/jobs2d/drivers/adapter/LineDrawerAdapter.java
```diff
@@ -0,0 +1,34 @@
...
+
+    @Override
+    public void setPosition(int x, int y) {
+        this.startX = x;
+        this.startY = y;
+    }
+
+    @Override
+    public void operateTo(int x, int y) {
+        ILine line = LineFactory.getSpecialLine();
```
rsusik, line 26: Unfortunately, this is not a generic solution (e.g. new line type requires new class, unnecessary direct coupling).

rsusik (commented):

PR #12, opened 2020-03-17
Skończone zadanie z Adapterem

snapeeek:
Maciej Woźniak 215925

### DIFF
--- a/src/main/java/edu/kis/powp/jobs2d/drivers/adapter/DrawingControllerAdapter.java
+++ b/src/main/java/edu/kis/powp/jobs2d/drivers/adapter/DrawingControllerAdapter.java
@@ -5,13 +5,17 @@ import edu.kis.legacy.drawer.shape.ILine;
 import edu.kis.legacy.drawer.shape.LineFactory;
 import edu.kis.powp.jobs2d.Job2dDriver;
 
+import javax.swing.*;
+
+import static edu.kis.powp.jobs2d.features.DrawerFeature.getDrawerController;
+
 /**
  * driver adapter to drawer with several bugs.
  */
-public class MyAdapter extends DrawPanelController implements Job2dDriver {
+public class DrawingControllerAdapter extends DrawPanelController implements Job2dDriver {
 	private int startX = 0, startY = 0;
 
-	public MyAdapter() {
+	public DrawingControllerAdapter() {
 		super();
 	}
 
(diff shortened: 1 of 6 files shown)


src/main/java/edu/kis/powp/jobs2d/drivers/adapter/DrawingControllerAdapter.java
```diff
@@ -5,13 +5,17 @@
 import edu.kis.legacy.drawer.shape.LineFactory;
 import edu.kis.powp.jobs2d.Job2dDriver;
 
+import javax.swing.*;
+
+import static edu.kis.powp.jobs2d.features.DrawerFeature.getDrawerController;
```
kowallus: Direct coupling with DrawerFeature class is unnecessary.

src/main/java/edu/kis/powp/jobs2d/drivers/adapter/DrawingControllerAdapter.java
```diff
@@ -5,13 +5,17 @@
...
+
 /**
  * driver adapter to drawer with several bugs.
  */
-public class MyAdapter extends DrawPanelController implements Job2dDriver {
+public class DrawingControllerAdapter extends DrawPanelController implements Job2dDriver {
```
kowallus: The name is better but it does not indicate the purpose of the adapter.

src/main/java/edu/kis/powp/jobs2d/drivers/adapter/SpecialLineAdapter.java
```diff
@@ -0,0 +1,44 @@
...
+import edu.kis.powp.jobs2d.Job2dDriver;
+
+import javax.swing.*;
+
+import static edu.kis.powp.jobs2d.features.DrawerFeature.getDrawerController;
+
+/**
+ * driver adapter to drawer with several bugs.
+ */
+public class SpecialLineAdapter extends DrawPanelController implements Job2dDriver {
```
kowallus: Similar issues

src/test/java/edu/kis/powp/jobs2d/events/SelectTestFigureOptionListener.java
```diff
@@ -9,13 +9,21 @@
...
+	{
 		this.driverManager = driverManager;
+		this.name = n;
 	}
 
 	@Override
 	public void actionPerformed(ActionEvent e) {
-		FiguresJoe.figureScript1(driverManager.getCurrentDriver());
+
+		if (name.equals("Joe1"))
```
kowallus: Enum or separate listeners would be better. At least you should consider constants.

src/main/java/edu/kis/powp/jobs2d/drivers/adapter/DrawingControllerAdapter.java
snapeeek: Inheritance is used once - in constructor.

PR #13, opened 2020-03-17
Adapter

AndrzejBilant:
AndrzejBilant

### DIFF
--- a/src/main/java/edu/kis/powp/jobs2d/drivers/adapter/DrawPanelControllerAdapter.java
+++ b/src/main/java/edu/kis/powp/jobs2d/drivers/adapter/DrawPanelControllerAdapter.java
@@ -5,13 +5,15 @@ import edu.kis.legacy.drawer.shape.ILine;
 import edu.kis.legacy.drawer.shape.LineFactory;
 import edu.kis.powp.jobs2d.Job2dDriver;
 
+import static edu.kis.powp.jobs2d.features.DrawerFeature.getDrawerController;
+
 /**
  * driver adapter to drawer with several bugs.
  */
-public class MyAdapter extends DrawPanelController implements Job2dDriver {
+public class DrawPanelControllerAdapter extends DrawPanelController implements Job2dDriver {
 	private int startX = 0, startY = 0;
 
-	public MyAdapter() {
+	public DrawPanelControllerAdapter() {
 		super();
 	}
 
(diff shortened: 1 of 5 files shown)


src/main/java/edu/kis/powp/jobs2d/drivers/adapter/DrawPanelControllerAdapter.java
```diff
@@ -5,13 +5,15 @@
 import edu.kis.legacy.drawer.shape.LineFactory;
 import edu.kis.powp.jobs2d.Job2dDriver;
 
+import static edu.kis.powp.jobs2d.features.DrawerFeature.getDrawerController;
+
 /**
  * driver adapter to drawer with several bugs.
  */
-public class MyAdapter extends DrawPanelController implements Job2dDriver {
+public class DrawPanelControllerAdapter extends DrawPanelController implements Job2dDriver {
```
kowallus: The name is better but it does not indicate the purpose of the adapter.

src/main/java/edu/kis/powp/jobs2d/drivers/adapter/DrawPanelControllerAdapter.java
```diff
@@ -5,13 +5,15 @@
 import edu.kis.legacy.drawer.shape.LineFactory;
 import edu.kis.powp.jobs2d.Job2dDriver;
 
+import static edu.kis.powp.jobs2d.features.DrawerFeature.getDrawerController;
```
kowallus: Direct coupling with DrawerFeature class is unnecessary.

src/main/java/edu/kis/powp/jobs2d/drivers/adapter/LineDrawerAdapter.java
```diff
@@ -0,0 +1,34 @@
+package edu.kis.powp.jobs2d.drivers.adapter;
+
+import edu.kis.legacy.drawer.panel.DrawPanelController;
+import edu.kis.legacy.drawer.shape.ILine;
+import edu.kis.legacy.drawer.shape.LineFactory;
+import edu.kis.powp.jobs2d.Job2dDriver;
+
+import static edu.kis.powp.jobs2d.features.DrawerFeature.getDrawerController;
+
+public class LineDrawerAdapter extends DrawPanelController implements Job2dDriver{
```
kowallus: Similar issues

src/test/java/edu/kis/powp/jobs2d/events/SelectTestFigureOptionListener.java
```diff
@@ -9,13 +9,15 @@
...
 	private DriverManager driverManager;
-
 	public SelectTestFigureOptionListener(DriverManager driverManager) {
 		this.driverManager = driverManager;
 	}
 
 	@Override
 	public void actionPerformed(ActionEvent e) {
-		FiguresJoe.figureScript1(driverManager.getCurrentDriver());
+		if(e.getActionCommand().contains("Figure Joe 1"))
```

rsusik:
I have nothing else to add, closing.

PR #14, opened 2020-03-19
Adapter finished

asmei1:


src/main/java/edu/kis/powp/jobs2d/drivers/adapter/DrawerAdapterToJobs2dDriver.java
```diff
@@ -6,13 +6,14 @@
 import edu.kis.powp.jobs2d.Job2dDriver;
 
 /**
- * driver adapter to drawer with several bugs.
+ * Driver adapter to drawer.
  */
-public class MyAdapter extends DrawPanelController implements Job2dDriver {
+public class DrawerAdapterToJobs2dDriver implements Job2dDriver {
```
kowallus: Not the other way around?  Jobs2dDriverToDrawerAdapter? 
BTW: The name does not need to mention the adapter pattern (it is not the design issue but only solution to the problem).

src/test/java/edu/kis/powp/jobs2d/TestJobs2dPatterns.java
```diff
@@ -83,7 +84,7 @@ public static void main(String[] args) {
 			public void run() {
 				Application app = new Application("2d jobs Visio");
 				DrawerFeature.setupDrawerPlugin(app);
-				setupDefaultDrawerVisibilityManagement(app);
+				//setupDefaultDrawerVisibilityManagement(app);
```
kowallus: It is safe to delete in revision repo

src/test/java/edu/kis/powp/jobs2d/events/SelectTestFigureOptionListener.java
```diff
@@ -7,15 +7,22 @@
...
 	public SelectTestFigureOptionListener(DriverManager driverManager) {
 		this.driverManager = driverManager;
 	}
-
+	
 	@Override
 	public void actionPerformed(ActionEvent e) {
-		FiguresJoe.figureScript1(driverManager.getCurrentDriver());
+		switch (e.getActionCommand()) {
+			case "Figure Joe 1":
```

kamsykula:
Should I fix this and make a new pull request or leave it like now?

src/main/java/edu/kis/powp/jobs2d/drivers/adapter/DrawerAdapterToJobs2dDriver.java
asmei1: Ok, now should be fixed.

src/test/java/edu/kis/powp/jobs2d/TestJobs2dPatterns.java
asmei1: Yes, I miss that.

src/test/java/edu/kis/powp/jobs2d/events/SelectTestFigureOptionListener.java
asmei1: Ok, I fixed this problem with enum

src/main/java/edu/kis/powp/jobs2d/drivers/adapter/DrawController.java
```diff
@@ -6,13 +6,14 @@
 import edu.kis.powp.jobs2d.Job2dDriver;
 
 /**
- * driver adapter to drawer with several bugs.
+ * Driver adapter to drawer.
  */
-public class MyAdapter extends DrawPanelController implements Job2dDriver {
+public class DrawController implements Job2dDriver {
```
rsusik, line 11: The name does not indicate the purpose of the adapter or what was actually adapted.

rsusik (commented):

PR #17, opened 2020-03-19
Laboratorium 1 - Adapter

krzysiek1105:


.idea/codeStyles/codeStyleConfig.xml
```diff
@@ -0,0 +1,5 @@
+<component name="ProjectCodeStyleConfiguration">
```
rsusik, line 1: Commits polluted with project files

src/main/java/edu/kis/powp/jobs2d/drivers/adapter/Jobs2dDriverToDrawerAdapter.java
```diff
@@ -0,0 +1,38 @@
...
+
+    @Override
+    public void setPosition(int x, int y) {
+        this.startX = x;
+        this.startY = y;
+    }
+
+    @Override
+    public void operateTo(int x, int y) {
+        ILine line = LineFactory.getBasicLine();
```
rsusik, line 26: Unfortunately, this is not a generic solution (e.g. new line type requires new class, unnecessary direct coupling).

src/test/java/edu/kis/powp/jobs2d/events/SelectTestFigureOptionListener.java
```diff
@@ -16,6 +16,13 @@ public SelectTestFigureOptionListener(DriverManager driverManager) {
...
 	@Override
 	public void actionPerformed(ActionEvent e) {
-		FiguresJoe.figureScript1(driverManager.getCurrentDriver());
+		switch(e.getActionCommand()) {
+			case "Figure Joe 1":
+				FiguresJoe.figureScript1(driverManager.getCurrentDriver());
+				break;
+			case "Figure Joe 2":
+				FiguresJoe.figureScript2(driverManager.getCurrentDriver());
+				break;
```
rsusik, line 25: Menu action should not depend on its caption. Enum or separate listeners would be better. At least you should consider constants.

rsusik (commented):

PR #18, opened 2020-03-19
Adapter

Lisek8:


src/main/java/edu/kis/powp/jobs2d/drivers/adapter/LineDrawerAdapter.java
```diff
@@ -0,0 +1,35 @@
...
+
+    @Override
+    public void setPosition(int x, int y) {
+        this.startX = x;
+        this.startY = y;
+    }
+
+    @Override
+    public void operateTo(int x, int y) {
+        ILine line = LineFactory.getSpecialLine();
```
rsusik, line 24: Unfortunately, this is not a generic solution (e.g. new line type requires new class, unnecessary direct coupling). Constructor injection should be used (using a parameter).

rsusik (commented):

PR #19, opened 2020-03-19
AdamWojtczak

AdamWojtczak:
numer indeksu: 215922

### DIFF
--- a/src/main/java/edu/kis/powp/jobs2d/drivers/adapter/DrawPanelAdapter.java
+++ b/src/main/java/edu/kis/powp/jobs2d/drivers/adapter/DrawPanelAdapter.java
@@ -5,13 +5,15 @@ import edu.kis.legacy.drawer.shape.ILine;
 import edu.kis.legacy.drawer.shape.LineFactory;
 import edu.kis.powp.jobs2d.Job2dDriver;
 
+import static edu.kis.powp.jobs2d.features.DrawerFeature.getDrawerController;
+
 /**
  * driver adapter to drawer with several bugs.
  */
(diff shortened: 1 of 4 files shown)


src/main/java/edu/kis/powp/jobs2d/drivers/adapter/SpecialLineDrawPanelAdapter.java
```diff
@@ -0,0 +1,38 @@
...
+
+    @Override
+    public void setPosition(int x, int y) {
+        this.startX = x;
+        this.startY = y;
+    }
+
+    @Override
+    public void operateTo(int x, int y) {
+        ILine line = LineFactory.getSpecialLine();
```
rsusik, line 25: Unfortunately, this is not a generic solution (e.g. new line type requires new class, unnecessary direct coupling).

src/main/java/edu/kis/powp/jobs2d/drivers/adapter/DrawPanelAdapter.java
```diff
@@ -5,14 +5,16 @@
...
  * driver adapter to drawer with several bugs.
  */
-public class MyAdapter extends DrawPanelController implements Job2dDriver {
+public class DrawPanelAdapter implements Job2dDriver {
 	private int startX = 0, startY = 0;
+	private DrawPanelController drawPanelController;
 
-	public MyAdapter() {
-		super();
+	public DrawPanelAdapter(DrawPanelController drawPanelController) {
```
rsusik, line 16: The name does not indicate the purpose of the adapter or what was actually adapted.

rsusik (commented):

PR #20, opened 2020-03-20
Zadanie Adapter

snapeeek:
Maciej Woźniak 215925


src/main/java/edu/kis/powp/jobs2d/drivers/adapter/DrawSpecialPanelAdapter.java
```diff
@@ -0,0 +1,41 @@
...
+    @Override
+    public void setPosition(int x, int y) {
+        this.startX = x;
+        this.startY = y;
+    }
+
+    @Override
+    public void operateTo(int x, int y) {
+        ILine line;
+        line = LineFactory.getSpecialLine();
```
rsusik, line 28: Unfortunately, this is not a generic solution, for each type of line another class is required.

rsusik (commented):
Duplicate of https://github.com/iis-powp-2020/powp_jobs2d/pull/12

PR #21, opened 2020-03-20
Adapter

MR-98:
Mateusz Rosiak 215866


src/main/java/edu/kis/powp/jobs2d/drivers/adapter/Jobs2dDriverToDrawerAdapter.java
```diff
@@ -0,0 +1,41 @@
...
+
+    @Override
+    public void setPosition(int x, int y) {
+        this.startX = x;
+        this.startY = y;
+    }
+
+    @Override
+    public void operateTo(int x, int y) {
+        ILine line = LineFactory.getBasicLine();
```
rsusik, line 29: Constructor injection should be used (using a parameter). This is unnecessary direct coupling with LineFactory class.

src/test/java/edu/kis/powp/jobs2d/events/SelectTestFigureOptionListener.java
```diff
@@ -8,14 +8,19 @@
...
-	public void actionPerformed(ActionEvent e) {
-		FiguresJoe.figureScript1(driverManager.getCurrentDriver());
-	}
+    @Override
+    public void actionPerformed(ActionEvent e) {
+        if (e.getActionCommand().toLowerCase().equals("figure joe 1")) {
+            FiguresJoe.figureScript1(driverManager.getCurrentDriver());
+        } else if (e.getActionCommand().toLowerCase().equals("figure joe 2")) {
+            FiguresJoe.figureScript2(driverManager.getCurrentDriver());
+        }
```
rsusik, line 23: Menu action should not depend on its caption. Enum or at least constants should be considered.

src/test/java/edu/kis/powp/jobs2d/events/SelectTestFigureOptionListener.java
```diff
@@ -8,14 +8,19 @@
...
 
-	private DriverManager driverManager;
+    private DriverManager driverManager;
 
-	public SelectTestFigureOptionListener(DriverManager driverManager) {
-		this.driverManager = driverManager;
-	}
+    public SelectTestFigureOptionListener(DriverManager driverManager) {
+        this.driverManager = driverManager;
+    }
```
rsusik, line 15: Formatting should follow the existing convention.

rsusik (commented):

PR #22, opened 2020-03-21
Adapter to point 3.4

rysiuziom1:


src/main/java/edu/kis/powp/jobs2d/drivers/adapter/LineDrawerAdapter.java
```diff
@@ -0,0 +1,42 @@
...
+        line.setStartCoordinates(this.startX, this.startY);
+        line.setEndCoordinates(x, y);
+
+        drawPanelController.drawLine(line);
+        setPosition(x, y);
+    }
+
+    @Override
+    public String toString() {
+        return "Buggy Simulator (" + line.getClass().getSimpleName() + ")";
```
rsusik: It is not buggy anymore.

rsusik (commented):

src/main/java/edu/kis/powp/jobs2d/drivers/adapter/LineDrawerAdapter.java
rysiuziom1: problem resolved in newest commit

PR #23, opened 2020-03-22
Adapter 3.2 & 3.3

Smirnoffq:


src/main/java/edu/kis/powp/jobs2d/drivers/adapter/LineDrawerAdapter.java
```diff
@@ -0,0 +1,38 @@
...
+
+    @Override
+    public void setPosition(int x, int y) {
+        this.startX = x;
+        this.startY = y;
+    }
+
+    @Override
+    public void operateTo(int x, int y) {
+        ILine line = LineFactory.getBasicLine();
```
rsusik, line 25: Unfortunately, this is not a generic solution (e.g. new line type requires new class, unnecessary direct coupling).

src/main/java/edu/kis/powp/jobs2d/drivers/adapter/DrawningPanelAdapter.java
```diff
@@ -8,11 +8,12 @@
 /**
  * driver adapter to drawer with several bugs.
  */
-public class MyAdapter extends DrawPanelController implements Job2dDriver {
+public class DrawningPanelAdapter implements Job2dDriver {
```
rsusik, line 11: The name does not indicate the purpose of the adapter or what was actually adapted.

src/test/java/edu/kis/powp/jobs2d/events/SelectTestFigureOptionListener.java
```diff
@@ -16,6 +16,13 @@ public SelectTestFigureOptionListener(DriverManager driverManager) {
...
 	public void actionPerformed(ActionEvent e) {
-		FiguresJoe.figureScript1(driverManager.getCurrentDriver());
+		switch(e.getActionCommand()) {
+			case "Figure Joe 1":
+				FiguresJoe.figureScript1(driverManager.getCurrentDriver());
+				break;
+			case "Figure Joe 2":
+				FiguresJoe.figureScript2(driverManager.getCurrentDriver());
+				break;
+		}
```
rsusik, line 26: Menu action should not depend on its caption. Enum or separate listeners would be better. At least you should consider constants.

rsusik (commented):

PR #24, opened 2020-03-22
Adapter 3.2 & 3.3

3elk4:
215911


src/test/java/edu/kis/powp/jobs2d/events/SelectTestFigureOptionListener.java
```diff
@@ -9,13 +9,25 @@
...
-		FiguresJoe.figureScript1(driverManager.getCurrentDriver());
+		switch(scriptNumber){
+			case 1:
+				FiguresJoe.figureScript1(driverManager.getCurrentDriver());
+				break;
+			case 2:
+				FiguresJoe.figureScript2(driverManager.getCurrentDriver());
+				break;
+			default:
+				break;
```
rsusik, line 29: Hardcoded values, enum or at least constants should be considered.

rsusik (commented):

PR #26, opened 2020-03-23
Adapter 3.2 and 3.3

Thomson19:


src/test/java/edu/kis/powp/jobs2d/events/SelectTestFigureOptionListener.java
```diff
@@ -16,6 +16,13 @@ public SelectTestFigureOptionListener(DriverManager driverManager) {
...
 	@Override
 	public void actionPerformed(ActionEvent e) {
-		FiguresJoe.figureScript1(driverManager.getCurrentDriver());
+		switch (e.getActionCommand()){
+			case "Figure Joe 1":
+				FiguresJoe.figureScript1(driverManager.getCurrentDriver());
+				break;
+			case "Figure Joe 2":
+				FiguresJoe.figureScript2(driverManager.getCurrentDriver());
+				break;
```
rsusik, line 25: Menu action should not depend on its caption. Enum or at least constants should be considered.

rsusik (commented):

PR #27, opened 2020-03-23
Adapter

215848:


src/main/java/edu/kis/powp/jobs2d/drivers/adapter/DrawPanelAdapter.java
```diff
@@ -2,16 +2,17 @@
...
 import edu.kis.legacy.drawer.shape.ILine;
+import static edu.kis.powp.jobs2d.features.DrawerFeature.getDrawerController;
 import edu.kis.legacy.drawer.shape.LineFactory;
 import edu.kis.powp.jobs2d.Job2dDriver;
 
 /**
  * driver adapter to drawer with several bugs.
  */
-public class MyAdapter extends DrawPanelController implements Job2dDriver {
+public class DrawPanelAdapter extends DrawPanelController implements Job2dDriver {
```
rsusik, line 12: The name does not indicate the purpose of the adapter or what was actually adapted.

src/test/java/edu/kis/powp/jobs2d/events/SelectTestFigureOptionListener.java
```diff
@@ -16,6 +16,10 @@ public SelectTestFigureOptionListener(DriverManager driverManager) {
 
 	@Override
 	public void actionPerformed(ActionEvent e) {
-		FiguresJoe.figureScript1(driverManager.getCurrentDriver());
+		if(e.getActionCommand().equals("Figure Joe 1"))
+			FiguresJoe.figureScript1(driverManager.getCurrentDriver());
+		else if(e.getActionCommand().equals("Figure Joe 2"))
+			FiguresJoe.figureScript2(driverManager.getCurrentDriver());
+
```
rsusik, line 23: Menu action should not depend on its caption. Enum or separate listeners would be better. At least you should consider constants.

rsusik (commented):

PR #30, opened 2020-03-24
Lab 2 - Adapter - 3.2 3.3 3.4

Kristrox:


src/test/java/edu/kis/powp/jobs2d/events/SelectTestFigureOptionListener.java
```diff
@@ -16,6 +18,18 @@ public SelectTestFigureOptionListener(DriverManager driverManager) {
...
+				break;
+			case "Figure Joe 2":
+				FiguresJoe.figureScript2(driverManager.getCurrentDriver());
+				break;
+			case "Figure Jane 1":
+				FiguresJane.figureScript(new DriverAdapter(driverManager.getCurrentDriver()));
+				break;
+			default:
+				System.out.println("Bad decision");
+		}
```
rsusik, line 33: Menu action should not depend on its caption. Enum or separate listeners would be better. At least you should consider constants.

src/main/java/edu/kis/powp/jobs2d/drivers/adapter/DrawPanelAdapter.java
```diff
@@ -27,11 +28,13 @@ public void operateTo(int x, int y) {
...
-		drawLine(line);
+		//drawLine(line);
+		setPosition(x,y);
+		DrawerFeature.getDrawerController().drawLine(line);
 	}
 
 	@Override
 	public String toString() {
-		return "@Q!$!@$!#@$(*#@&Q(%^*#@";
+		return "Buggy Simulator";
```
rsusik, line 38: It is not buggy now, right?

rsusik (commented):

PR #31, opened 2020-03-24
Adapter do 3.3

Lefiuu:


.idea/compiler.xml
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8"?>
```

src/main/java/edu/kis/powp/jobs2d/drivers/adapter/LineDrawerAdapter.java
```diff
@@ -0,0 +1,40 @@
...
+
+    @Override
+    public void setPosition(int x, int y) {
+        this.startX = x;
+        this.startY = y;
+    }
+
+    @Override
+    public void operateTo(int x, int y) {
+        ILine line = LineFactory.getSpecialLine();
```
rsusik, line 26: Unfortunately, this is not a generic solution (e.g. new line type requires new class, unnecessary direct coupling).

src/test/java/edu/kis/powp/jobs2d/events/SelectTestFigureOptionListener.java
```diff
@@ -16,6 +16,10 @@ public SelectTestFigureOptionListener(DriverManager driverManager) {
 
 	@Override
 	public void actionPerformed(ActionEvent e) {
-		FiguresJoe.figureScript1(driverManager.getCurrentDriver());
+
+		if(e.getActionCommand().equals("Figure Joe 1"))
+			FiguresJoe.figureScript1(driverManager.getCurrentDriver());
+		else if( e.getActionCommand().equals("Figure Joe 2"))
+			FiguresJoe.figureScript2(driverManager.getCurrentDriver());
```
rsusik, line 23: Menu action should not depend on its caption. Enum or at least constants should be considered.

rsusik (commented):

PR #33, opened 2020-03-24
Adapter zad 3.2 & 3.3

Trackres:


jobs2d.iml
```diff
@@ -0,0 +1,41 @@
...
+      <library name="Maven: edu.kis.powp:appbase:1.0.0">
+        <CLASSES>
+          <root url="jar://$MODULE_DIR$/lib/PowpAppBase.jar!/" />
+        </CLASSES>
+        <JAVADOC />
+        <SOURCES />
+      </library>
+    </orderEntry>
+  </component>
+</module>
```
rsusik, line 41: Commits polluted with project files

src/main/java/edu/kis/powp/jobs2d/drivers/adapter/DrawPanelAdapter.java
```diff
@@ -26,12 +33,13 @@ public void operateTo(int x, int y) {
 		ILine line = LineFactory.getBasicLine();
```
rsusik, line 33: Unfortunately, this is not a generic solution (e.g. new line type requires new class).

src/main/java/edu/kis/powp/jobs2d/drivers/adapter/DrawPanelAdapter.java
```diff
@@ -4,17 +4,24 @@
...
+import edu.kis.powp.jobs2d.features.DrawerFeature;
+import edu.kis.powp.jobs2d.magicpresets.FiguresJoe;
+
+import static edu.kis.powp.jobs2d.features.DrawerFeature.getDrawerController;
 
 /**
  * driver adapter to drawer with several bugs.
  */
-public class MyAdapter extends DrawPanelController implements Job2dDriver {
+public class DrawPanelAdapter extends DrawPanelController implements Job2dDriver {
```
rsusik, line 15: The name does not indicate the purpose of the adapter or what was actually adapted.

src/test/java/edu/kis/powp/jobs2d/events/SelectTestFigureOptionListener.java
```diff
@@ -16,6 +16,13 @@ public SelectTestFigureOptionListener(DriverManager driverManager) {
...
 	public void actionPerformed(ActionEvent e) {
-		FiguresJoe.figureScript1(driverManager.getCurrentDriver());
+		switch (e.getActionCommand()){
+			case "Figure Joe 1":
+				FiguresJoe.figureScript1(driverManager.getCurrentDriver());
+				break;
+			case "Figure Joe 2":
+				FiguresJoe.figureScript2(driverManager.getCurrentDriver());
+				break;
+		}
```
rsusik, line 26: Menu action should not depend on its caption. Enum or separate listeners would be better. At least you should consider constants.

rsusik (commented):

PR #35, opened 2020-03-24
Task 3.2 & 3.3 & 3.4

k-chr:
Student: Kamil Chrustowski 
Nr Indeksu: 215691

### DIFF
--- a/src/main/java/edu/kis/powp/jobs2d/drivers/DriverManager.java
+++ b/src/main/java/edu/kis/powp/jobs2d/drivers/DriverManager.java
@@ -9,19 +9,19 @@ import edu.kis.powp.jobs2d.LoggerDriver;
  */
 public class DriverManager {
 
-	private Job2dDriver currentDriver = new LoggerDriver();
+    private Job2dDriver currentDriver = new LoggerDriver();
 
-	/**
-	 * @param driver Set the driver as current.
-	 */
-	public synchronized void setCurrentDriver(Job2dDriver driver) {
-		currentDriver = driver;
-	}
+    /**
+     * @return Current driver.
+     */
+    public synchronized Job2dDriver getCurrentDriver() {
+        return currentDriver;
+    }
 
-	/**
-	 * @return Current driver.
-	 */
-	public synchronized Job2dDriver getCurrentDriver() {
-		return currentDriver;
-	}
+    /**
+     * @param driver Set the driver as current.
+     */
+    public synchronized void setCurrentDriver(Job2dDriver driver) {
+        currentDriver = driver;
+    }
 }
--- a/src/main/java/edu/kis/powp/jobs2d/drivers/adapter/Job2dDriverAdapter.java
+++ b/src/main/java/edu/kis/powp/jobs2d/drivers/adapter/Job2dDriverAdapter.java
@@ -11,31 +11,32 @@ import javax.swing.*;
  * driver adapter to drawer with several bugs.
  */
 public class Job2dDriverAdapter implements Job2dDriver {
-	private int startX = 0, startY = 0;
-	private DrawPanelController drawPanelController;
 
-	public Job2dDriverAdapter(DrawPanelController drawPanelController) {
-		this.drawPanelController = drawPanelController;
-	}
+    private int startX = 0, startY = 0;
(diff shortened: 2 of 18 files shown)


src/main/java/edu/kis/powp/jobs2d/drivers/DriverManager.java
```diff
@@ -9,19 +9,19 @@
...
-	 */
-	public synchronized Job2dDriver getCurrentDriver() {
-		return currentDriver;
-	}
+    /**
+     * @param driver Set the driver as current.
+     */
+    public synchronized void setCurrentDriver(Job2dDriver driver) {
+        currentDriver = driver;
+    }
```
rsusik, line 26: Something bad happened here

src/main/java/edu/kis/powp/jobs2d/drivers/adapter/Job2dDriverAdapter.java
```diff
@@ -0,0 +1,42 @@
...
+
+    @Override
+    public void setPosition(int x, int y) {
+        this.startX = x;
+        this.startY = y;
+    }
+
+    @Override
+    public void operateTo(int x, int y) {
+        ILine line = LineFactory.getBasicLine();
```
rsusik, line 30: Direct coupling with LineFactory class is unnecessary. You should consider constructor injection (using a parameter).

rsusik (commented):
Note: You should keep the original formatting of code you are changing. Usually commits that do not follow the formatting rules are rejected.

src/main/java/edu/kis/powp/jobs2d/drivers/adapter/LineDrawerAdapter.java
```diff
@@ -0,0 +1,61 @@
...
+    private int x = 0;
+    private int y = 0;
+    private ILine lineToDraw = null;
+
+    public LineDrawerAdapter(DrawPanelController controller, ILine lineToDraw) {
+        this.controller = controller;
+        this.lineToDraw = lineToDraw;
+    }
+
+    public LineDrawerAdapter(DrawPanelController controller, LineType type) {
```
kowallus, line 21: I would recommend simpler solutions (KISS principle ;) ). The first constructor is sufficient. This one is redundant and introduces new dependencies.

src/main/java/edu/kis/powp/jobs2d/shapes/lines/UserModifiableLine.java
```diff
@@ -0,0 +1,62 @@
...
+        this.color = DEFAULT_COLOR;
+        this.dotted = DEFAULT_STYLE;
+    }
+
+    public static LineBuilder builder() {
+        return new LineBuilder();
+    }
+
+    public LineBuilder rebuilder() {
+        return new LineBuilder();
```
kowallus, line 30: It does not differ from a static builder() method. If it could pass current line params it would make sense (similar to instance cloning).

src/main/java/edu/kis/powp/jobs2d/drivers/adapter/Job2dDriverAdapter.java
k-chr: Fixed in latest commit to master branch.

src/main/java/edu/kis/powp/jobs2d/drivers/adapter/LineDrawerAdapter.java
k-chr: Fixed in latest commit to master branch.

src/main/java/edu/kis/powp/jobs2d/shapes/lines/UserModifiableLine.java
k-chr: Fixed in latest commit to master branch.

src/main/java/edu/kis/powp/jobs2d/drivers/DriverManager.java
k-chr: Ok, I understood the problem, on the other hand I think, that including formatting provider in xml file (that can be imported into IDE and used by every developer working on this project) should be good solution to avoid issues like this above.

PR #36, opened 2020-03-24
Task 3.2 and 3.3

Luvkitri:


.idea/compiler.xml
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8"?>
```

src/main/java/edu/kis/powp/jobs2d/drivers/adapter/PanelDrawerAdapter.java
```diff
@@ -4,15 +4,19 @@
...
 
 /**
  * driver adapter to drawer with several bugs.
  */
-public class MyAdapter extends DrawPanelController implements Job2dDriver {
-	private int startX = 0, startY = 0;
 
-	public MyAdapter() {
-		super();
+public class PanelDrawerAdapter extends DrawerFeature implements Job2dDriver {
```
rsusik, line 13: Unused code. Should be removed.

rsusik (commented):

PR #38, opened 2020-03-24
Lab1 - Adapter finish

Paulina-Koska:


src/main/java/edu/kis/powp/jobs2d/drivers/adapter/LineDrawerAdapter.java
```diff
@@ -0,0 +1,37 @@
...
+        line.setStartCoordinates(this.startX, this.startY);
+        line.setEndCoordinates(x, y);
+        setPosition(x, y);
+
+        drawPanelController.drawLine(line);
+    }
+
+    @Override
+    public String toString() {
+        return "Buggy Simulator 2";
```
rsusik, line 35: Is it buggy?

src/test/java/edu/kis/powp/jobs2d/events/SelectTestFigureOptionListener.java
```diff
@@ -16,6 +16,9 @@ public SelectTestFigureOptionListener(DriverManager driverManager) {
 
 	@Override
 	public void actionPerformed(ActionEvent e) {
-		FiguresJoe.figureScript1(driverManager.getCurrentDriver());
+		if(e.getActionCommand().equals("Figure Joe 1"))
+			FiguresJoe.figureScript1(driverManager.getCurrentDriver());
+		else
+			FiguresJoe.figureScript2(driverManager.getCurrentDriver());
```
rsusik, line 22: Menu action should not depend on its caption. Enum or separate listeners would be better. At least you should consider constants.

rsusik (commented):

PR #39, opened 2020-03-24
Adapter 3.2 and 3.3

startled-cat:


jobs2d.iml
```diff
@@ -0,0 +1,44 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
rsusik, line 1: Commits polluted with project files

src/main/java/edu/kis/powp/jobs2d/drivers/adapter/DrawDriver.java
```diff
@@ -8,11 +8,14 @@
 /**
  * driver adapter to drawer with several bugs.
  */
-public class MyAdapter extends DrawPanelController implements Job2dDriver {
+public class DrawDriver implements Job2dDriver {
```
rsusik, line 11: The name does not indicate the purpose of the adapter or what was actually adapted.

src/test/java/edu/kis/powp/jobs2d/events/SelectTestFigureOptionListener.java
```diff
@@ -9,13 +9,36 @@
...
+		this.driverManager = driverManager;
+		this.testType = testType;
 	}
 
 	@Override
 	public void actionPerformed(ActionEvent e) {
-		FiguresJoe.figureScript1(driverManager.getCurrentDriver());
+
+		switch(testType){
+			case 1:
```
rsusik, line 28: Hardcoded value, enum should be used.

rsusik (commented):

PR #40, opened 2020-03-24
Adapter - zadania 3.2 i 3.3

BSkura98:
Indeks: 215877

### DIFF
--- a/src/test/java/edu/kis/powp/jobs2d/events/SelectTestFigure1OptionListener.java
+++ b/src/test/java/edu/kis/powp/jobs2d/events/SelectTestFigure1OptionListener.java
@@ -6,11 +6,11 @@ import java.awt.event.ActionListener;
 import edu.kis.powp.jobs2d.drivers.DriverManager;
 import edu.kis.powp.jobs2d.magicpresets.FiguresJoe;
 
-public class SelectTestFigureOptionListener implements ActionListener {
+public class SelectTestFigure1OptionListener implements ActionListener {
 
 	private DriverManager driverManager;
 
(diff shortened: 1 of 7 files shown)


src/test/java/edu/kis/powp/jobs2d/events/SelectTestFigure1OptionListener.java
```diff
@@ -6,11 +6,11 @@
 import edu.kis.powp.jobs2d.drivers.DriverManager;
 import edu.kis.powp.jobs2d.magicpresets.FiguresJoe;
 
-public class SelectTestFigureOptionListener implements ActionListener {
+public class SelectTestFigure1OptionListener implements ActionListener {
```
rsusik, line 9: Unfortunately, this is not a generic solution, for each figure another class is required.

src/test/java/edu/kis/powp/jobs2d/events/SelectTestFigure2OptionListener.java
```diff
@@ -0,0 +1,21 @@
+package edu.kis.powp.jobs2d.events;
+
+import java.awt.event.ActionEvent;
+import java.awt.event.ActionListener;
+
+import edu.kis.powp.jobs2d.drivers.DriverManager;
+import edu.kis.powp.jobs2d.magicpresets.FiguresJoe;
+
+public class SelectTestFigure2OptionListener implements ActionListener {
```
rsusik, line 9: Same here

rsusik (commented):

PR #43, opened 2020-03-25
Adapter zadanie 3.2.1

bopke:


src/main/java/edu/kis/powp/jobs2d/drivers/adapter/ScreenDrawerDriver.java
```diff
@@ -0,0 +1,38 @@
...
+        this.startY = y;
+    }
+
+    @Override
+    public void operateTo(int x, int y) {
+        ILine line = LineFactory.getBasicLine();
+        line.setStartCoordinates(this.startX, this.startY);
+        line.setEndCoordinates(x, y);
+        this.startX = x;
+        this.startY = y;
```
kowallus, line 30: Minor code redundancy

PR #44, opened 2020-03-25
3.4.1 i 3.4.2

maybegamer:
215693


src/main/java/edu/kis/powp/jobs2d/drivers/adapter/LineDrawerAdapter.java
```diff
@@ -0,0 +1,38 @@
...
+    public LineDrawerAdapter(DrawPanelController drawPanelController, ILine line) {
+        this.drawPanelController = drawPanelController;
+        this.line = line;
+    }
+
+    @Override
+    public void setPosition(int x, int y) {
+        this.startX = x;
+        this.startY = y;
+        this.startY = y;
```
kowallus, line 21: it happens

kowallus (commented):
5.5 pts

A minor mistake in UML: aggregation on the wrong side.

PR #46, opened 2020-03-25
215719

spacja255:
Przemysław Frąg
Numer indeksu: 215719
Grupa: IO1
Zadania: 3.1, 3.2


src/main/java/edu/kis/powp/jobs2d/drivers/adapter/DrawDriver.java
```diff
@@ -27,7 +23,8 @@ public void operateTo(int x, int y) {
 		line.setStartCoordinates(this.startX, this.startY);
 		line.setEndCoordinates(x, y);
 
-		drawLine(line);
+		DrawerFeature.getDrawerController().drawLine(line);
```
kowallus, line 26: This is unnecessary direct coupling with DrawerFeature class.

kowallus (commented):
Only one remark. OK!

PR #49, opened 2020-03-26
Adapter

PrzemaXster:


jobs2d.iml
```diff
@@ -0,0 +1,44 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
kowallus, line 1: Commits polluted with project files

src/main/java/edu/kis/powp/jobs2d/drivers/adapter/StraightLineAdapter.java
```diff
@@ -1,18 +1,20 @@
...
 import edu.kis.legacy.drawer.shape.ILine;
 import edu.kis.legacy.drawer.shape.LineFactory;
 import edu.kis.powp.jobs2d.Job2dDriver;
+import edu.kis.powp.jobs2d.features.DrawerFeature;
 
 /**
  * driver adapter to drawer with several bugs.
  */
-public class MyAdapter extends DrawPanelController implements Job2dDriver {
+public class StraightLineAdapter implements Job2dDriver {
```
kowallus, line 12: The name is better but it lacks a driver aspect.

src/main/java/edu/kis/powp/jobs2d/drivers/adapter/StraightLineAdapter.java
```diff
@@ -1,18 +1,20 @@
...
 	private int startX = 0, startY = 0;
-
-	public MyAdapter() {
+	private DrawPanelController drawController;
+	public StraightLineAdapter() {
 		super();
+		drawController = DrawerFeature.getDrawerController();
```
kowallus, line 17: Constructor injection should be used (using a parameter). This is unnecessary direct coupling with DrawerFeature class.

src/main/java/edu/kis/powp/jobs2d/drivers/adapter/StraightLineAdapter.java
```diff
@@ -23,15 +25,17 @@ public void setPosition(int x, int y) {
...
 	@Override
 	public void operateTo(int x, int y) {
+
 		ILine line = LineFactory.getBasicLine();
 		line.setStartCoordinates(this.startX, this.startY);
 		line.setEndCoordinates(x, y);
-
-		drawLine(line);
+		this.startX = x;
+		this.startY = y;
```
kowallus, line 33: Minor code duplication

kowallus (commented):
4 pts

UML diagram: 
- aggregation on the wrong side of the relation.
- no multiplicities

PR #51, opened 2020-03-27
Adapter

zenakz:
215934


src/test/java/edu/kis/powp/jobs2d/events/SelectTestFigureOptionListener.java
```diff
@@ -16,6 +17,8 @@ public SelectTestFigureOptionListener(DriverManager driverManager) {
 
 	@Override
 	public void actionPerformed(ActionEvent e) {
-		FiguresJoe.figureScript1(driverManager.getCurrentDriver());
+		DrawerFeature.getDrawerController().clearPanel();
+		if(e.getActionCommand().equals("Figure Joe 2"))FiguresJoe.figureScript2(driverManager.getCurrentDriver());
+		else FiguresJoe.figureScript1(driverManager.getCurrentDriver());
```
rsusik, line 22: Menu action should not depend on its caption. Enum or at least constants should be considered.

rsusik (commented):

PR #52, opened 2020-03-28
Version after reforking with 3.2 & 3.3

Krizuu:
After second fork
215668


.idea/compiler.xml
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8"?>
```

src/main/java/edu/kis/powp/jobs2d/drivers/adapter/DrawPanelAdapter.java
```diff
@@ -0,0 +1,45 @@
...
+
+	@Override
+	public void setPosition(int x, int y) {
+		this.startX = x;
+		this.startY = y;
+	}
+
+	@Override
+	public void operateTo(int x, int y) {
+		ILine line = LineFactory.getBasicLine();
```
rsusik, line 33: Constructor injection should be used (using a parameter). This is unnecessary direct coupling with LineFactory class.

src/test/java/edu/kis/powp/jobs2d/events/SelectTestFigureOptionListener.java
```diff
@@ -16,6 +16,13 @@ public SelectTestFigureOptionListener(DriverManager driverManager) {
...
 	@Override
 	public void actionPerformed(ActionEvent e) {
-		FiguresJoe.figureScript1(driverManager.getCurrentDriver());
+		switch (e.getActionCommand()){
+			case "Figure Joe 1":
+				FiguresJoe.figureScript1(driverManager.getCurrentDriver());
+				break;
+			case "Figure Joe 2":
+				FiguresJoe.figureScript2(driverManager.getCurrentDriver());
+				break;
```
rsusik, line 25: Menu action should not depend on its caption. Enum or at least constants should be considered.

rsusik (commented):

PR #53, opened 2020-03-28
fixed drawer

matlukomski94:


rsusik (commented):
3.2 is partially completed (only one figure).
3.3 completed, but only one "special" line type is available.

PR #56, opened 2020-03-30
3.2, 3.3

tmajcher:
215810 Tomasz Majcher


.idea/codeStyles/codeStyleConfig.xml
```diff
@@ -0,0 +1,5 @@
+<component name="ProjectCodeStyleConfiguration">
```

src/main/java/edu/kis/powp/jobs2d/drivers/adapter/DrawerAdapter.java
```diff
@@ -4,14 +4,15 @@
 import edu.kis.legacy.drawer.shape.ILine;
 import edu.kis.legacy.drawer.shape.LineFactory;
 import edu.kis.powp.jobs2d.Job2dDriver;
+import edu.kis.powp.jobs2d.features.DrawerFeature;
 
 /**
  * driver adapter to drawer with several bugs.
  */
-public class MyAdapter extends DrawPanelController implements Job2dDriver {
+public class DrawerAdapter extends DrawPanelController implements Job2dDriver {
```
kowallus, line 12: The name is better but it lacks a driver aspect.

src/main/java/edu/kis/powp/jobs2d/drivers/adapter/DrawerAdapter.java
```diff
@@ -26,8 +27,10 @@ public void operateTo(int x, int y) {
 		ILine line = LineFactory.getBasicLine();
 		line.setStartCoordinates(this.startX, this.startY);
 		line.setEndCoordinates(x, y);
+		setPosition(x,y);
 
-		drawLine(line);
+		DrawerFeature.getDrawerController().drawLine(line);
+//		drawLine(line);
```
kowallus, line 33: In revision repo it is safe to delete code instead of using a comment.

src/main/java/edu/kis/powp/jobs2d/drivers/adapter/DrawerAdapter.java
```diff
@@ -26,8 +27,10 @@ public void operateTo(int x, int y) {
 		ILine line = LineFactory.getBasicLine();
 		line.setStartCoordinates(this.startX, this.startY);
 		line.setEndCoordinates(x, y);
+		setPosition(x,y);
 
-		drawLine(line);
+		DrawerFeature.getDrawerController().drawLine(line);
```
kowallus, line 32: Constructor injection should be used (using a parameter). This is unnecessary direct coupling with DrawerFeature class.

src/main/java/edu/kis/powp/jobs2d/drivers/adapter/LineDrawerAdapter.java
```diff
@@ -0,0 +1,37 @@
...
+    @Override
+    public void operateTo(int x, int y) {
+        iLine.setStartCoordinates(this.startX, this.startY);
+        iLine.setEndCoordinates(x, y);
+
+
+        DrawerFeature.getDrawerController().drawLine(iLine);
+        setPosition(x,y);
+    }
+
```
kowallus, line 32: same issues

src/test/java/edu/kis/powp/jobs2d/events/SelectTestFigureOptionListener.java
```diff
@@ -17,5 +17,15 @@ public SelectTestFigureOptionListener(DriverManager driverManager) {
 	@Override
 	public void actionPerformed(ActionEvent e) {
 		FiguresJoe.figureScript1(driverManager.getCurrentDriver());
+		switch(e.getActionCommand()){
+			case "Figure Joe 1":
```

kowallus (commented):
3 pts (lab-adapter)

PR #57, opened 2020-03-30
Pull request z kompletem zadań

Kacpri:


src/test/java/edu/kis/powp/jobs2d/events/SelectTestFigureOptionListener.java
```diff
@@ -16,6 +18,16 @@ public SelectTestFigureOptionListener(DriverManager driverManager) {
 
 	@Override
 	public void actionPerformed(ActionEvent e) {
-		FiguresJoe.figureScript1(driverManager.getCurrentDriver());
+		switch (e.getActionCommand()) {
+			case "Figure Joe 1":
```
kowallus, line 22: Menu action should not depend on its caption. Enum or separate listeners would be better. At least you should consider constants.

src/main/java/edu/kis/powp/jobs2d/drivers/adapter/DrawPanelAdapter.java
```diff
@@ -8,11 +8,12 @@
 /**
  * driver adapter to drawer with several bugs.
  */
-public class MyAdapter extends DrawPanelController implements Job2dDriver {
+public class DrawPanelAdapter implements Job2dDriver {
```
kowallus, line 11: The name is better but it lacks a driver aspect.

kowallus:
5 pts

Minor mistakes in UML: 
- multiplicity 1..* on the side of DrawPanelController.
- aggregation on the wrong side.

PR #58, opened 2020-03-30
Adapter 3.2 3.3 3.4

arekdziuda:
Arkadiusz Dziuda
215714


.idea/codeStyles/codeStyleConfig.xml
```diff
@@ -0,0 +1,5 @@
+<component name="ProjectCodeStyleConfiguration">
```

src/main/java/edu/kis/powp/jobs2d/drivers/adapter/DrawerAdapter.java
```diff
@@ -4,14 +4,15 @@
 import edu.kis.legacy.drawer.shape.ILine;
 import edu.kis.legacy.drawer.shape.LineFactory;
 import edu.kis.powp.jobs2d.Job2dDriver;
+import edu.kis.powp.jobs2d.features.DrawerFeature;
 
 /**
  * driver adapter to drawer with several bugs.
  */
-public class MyAdapter extends DrawPanelController implements Job2dDriver {
+public class DrawerAdapter extends DrawPanelController implements Job2dDriver {
```
kowallus, line 12: Inheritance is not used (no super drawLine calls). Should be removed!

src/main/java/edu/kis/powp/jobs2d/drivers/adapter/DrawerAdapter.java
```diff
@@ -26,8 +27,9 @@ public void operateTo(int x, int y) {
 		ILine line = LineFactory.getBasicLine();
 		line.setStartCoordinates(this.startX, this.startY);
 		line.setEndCoordinates(x, y);
+		setPosition(x, y);
 
-		drawLine(line);
+		DrawerFeature.getDrawerController().drawLine(line);
```
kowallus, line 32: Constructor injection should be used (using a parameter). This is unnecessary direct coupling with DrawerFeature class.

src/main/java/edu/kis/powp/jobs2d/drivers/adapter/LineDrawerAdapter.java
```diff
@@ -0,0 +1,58 @@
...
+    private int x = 0;
+    private int y = 0;
+    private ILine lineToDraw = null;
+
+    public LineDrawerAdapter(DrawPanelController controller, LineType type){
+        this.controller = controller;
+
+        switch(type){
+            case BASIC:
+                this.lineToDraw = LineFactory.getBasicLine();
```
kowallus, line 25: Unnecessary coupling with LineFactory and LineType (KISS principle).

src/test/java/edu/kis/powp/jobs2d/events/SelectTestFigureOptionListener.java
```diff
@@ -16,6 +18,13 @@ public SelectTestFigureOptionListener(DriverManager driverManager) {
 
 	@Override
 	public void actionPerformed(ActionEvent e) {
-		FiguresJoe.figureScript1(driverManager.getCurrentDriver());
+
+			if (e.getActionCommand().equals("Figure Joe 1"))
```
kowallus, line 22: Menu action should not depend on its caption. Enum or separate listeners would be better. At least you should consider constants.

kowallus (commented):
3.5 pts

PR #60, opened 2020-03-30
Adapter 3.2,3.3,3.4

DeanMiles:
Alan Franas 215717 grupa IO1


src/main/java/edu/kis/powp/jobs2d/drivers/adapter/DrawerAdapter.java
```diff
@@ -4,14 +4,15 @@
 import edu.kis.legacy.drawer.shape.ILine;
 import edu.kis.legacy.drawer.shape.LineFactory;
 import edu.kis.powp.jobs2d.Job2dDriver;
+import edu.kis.powp.jobs2d.features.DrawerFeature;
 
 /**
  * driver adapter to drawer with several bugs.
  */
-public class MyAdapter extends DrawPanelController implements Job2dDriver {
+public class DrawerAdapter extends DrawPanelController implements Job2dDriver {
```
kowallus, line 12: The name is better but it lacks a driver aspect.

src/main/java/edu/kis/powp/jobs2d/drivers/adapter/DrawerAdapter.java
```diff
@@ -26,8 +27,8 @@ public void operateTo(int x, int y) {
 		ILine line = LineFactory.getBasicLine();
 		line.setStartCoordinates(this.startX, this.startY);
 		line.setEndCoordinates(x, y);
-
-		drawLine(line);
+		setPosition(x, y);
+		DrawerFeature.getDrawerController().drawLine(line);
```
kowallus, line 31: Constructor injection should be used (using a parameter). This is unnecessary direct coupling with DrawerFeature class.

src/main/java/edu/kis/powp/jobs2d/drivers/adapter/LineDrawerAdapter.java
```diff
@@ -0,0 +1,37 @@
...
+    }
+
+    @Override
+    public void operateTo(int x, int y) {
+        iLine.setStartCoordinates(this.startX, this.startY);
+        iLine.setEndCoordinates(x, y);
+        DrawerFeature.getDrawerController().drawLine(iLine);
+        setPosition(x, y);
+    }
+}
```
kowallus, line 37: similar issues

kowallus (commented):
4 pts

UML diagram does not show the relation with DrawerFeature.

PR #61, opened 2020-03-31
Lab 1 Adapter

Radko-A:


src/main/java/edu/kis/powp/jobs2d/drivers/adapter/DrawerAdapter.java
```diff
@@ -27,7 +28,13 @@ public void operateTo(int x, int y) {
 		line.setStartCoordinates(this.startX, this.startY);
 		line.setEndCoordinates(x, y);
 
+		startX = x;
+		startY = y;
+
 		drawLine(line);
+		DrawerFeature.getDrawerController().drawLine(line);
```
kowallus, line 35: Constructor injection should be used (using a parameter). This is unnecessary direct coupling with DrawerFeature class.

src/main/java/edu/kis/powp/jobs2d/drivers/adapter/LineDrawerAdapter.java
```diff
@@ -0,0 +1,40 @@
...
+import edu.kis.legacy.drawer.panel.DrawPanelController;
+import edu.kis.legacy.drawer.shape.ILine;
+import edu.kis.legacy.drawer.shape.LineFactory;
+import edu.kis.powp.jobs2d.Job2dDriver;
+import edu.kis.powp.jobs2d.features.DrawerFeature;
+
+/**
+ * driver adapter to drawer with several bugs.
+ */
+public class LineDrawerAdapter extends DrawPanelController implements Job2dDriver {
```
kowallus, line 12: similar issues in this class

src/main/java/edu/kis/powp/jobs2d/drivers/adapter/LineDrawerAdapter.java
```diff
@@ -0,0 +1,40 @@
...
+
+    @Override
+    public void setPosition(int x, int y) {
+        this.startX = x;
+        this.startY = y;
+    }
+
+    @Override
+    public void operateTo(int x, int y) {
+        ILine line = LineFactory.getSpecialLine();
```
kowallus, line 27: Hardcoded line type. What about 3.3.2 test?

Unfortunately, this is not a generic solution. (e.g. new line type requires new class, unnecessary direct coupling)

src/test/java/edu/kis/powp/jobs2d/events/SelectTestFigureOptionListener.java
```diff
@@ -16,6 +16,11 @@ public SelectTestFigureOptionListener(DriverManager driverManager) {
 
 	@Override
 	public void actionPerformed(ActionEvent e) {
-		FiguresJoe.figureScript1(driverManager.getCurrentDriver());
+
+		if(e.getActionCommand().contains("Figure Joe 1")) {
```

kowallus (commented):
2.5 pts

UML diagram missing

PR #64, opened 2020-04-01
Lab 3.3

GGuinea:
Kamil Zielinski
215933


src/main/java/edu/kis/powp/jobs2d/drivers/adapter/DrawController.java
```diff
@@ -8,11 +8,13 @@
 /**
  * driver adapter to drawer with several bugs.
  */
-public class MyAdapter extends DrawPanelController implements Job2dDriver {
+public class DrawController implements Job2dDriver {
```
kowallus, line 11: The name is better but it lacks a driver aspect.

kowallus (commented):
5 pts

UML diagram is missing

PR #65, opened 2020-04-01
Adapter 3.2 3.3

pawellazicki:


.idea/codeStyles/codeStyleConfig.xml
```diff
@@ -0,0 +1,5 @@
+<component name="ProjectCodeStyleConfiguration">
```

src/main/java/edu/kis/powp/jobs2d/drivers/adapter/DrawPanelAdapter.java
```diff
@@ -5,14 +5,12 @@
 import edu.kis.legacy.drawer.shape.LineFactory;
 import edu.kis.powp.jobs2d.Job2dDriver;
 
-/**
- * driver adapter to drawer with several bugs.
- */
-public class MyAdapter extends DrawPanelController implements Job2dDriver {
+public class DrawPanelAdapter extends DrawPanelController implements Job2dDriver {
```
kowallus, line 8: Inheritance is not used (no super drawLine calls). Should be removed!

src/test/java/edu/kis/powp/jobs2d/events/SelectTestFigureOptionListener.java
```diff
@@ -16,6 +16,10 @@ public SelectTestFigureOptionListener(DriverManager driverManager) {
 
 	@Override
 	public void actionPerformed(ActionEvent e) {
-		FiguresJoe.figureScript1(driverManager.getCurrentDriver());
+		if (e.getActionCommand().toLowerCase().equals("figure joe 1")) {
```

kowallus (commented):
4 pts

On UML:
- directed association is mistaken for inheritance.
- there lacks an (still) existing inheritance between adapter and DrawPanelController.

PR #66, opened 2020-04-01
Lab Adapter 

Arrlen:


src/main/java/edu/kis/powp/jobs2d/drivers/adapter/DrawingAdapterDriver.java
```diff
@@ -26,12 +28,13 @@ public void operateTo(int x, int y) {
 		ILine line = LineFactory.getBasicLine();
 		line.setStartCoordinates(this.startX, this.startY);
 		line.setEndCoordinates(x, y);
-
-		drawLine(line);
+		this.startX = x;
+		this.startY = y;
+		DrawerFeature.getDrawerController().drawLine(line);
```
kowallus, line 33: Constructor injection should be used (using a parameter). This is unnecessary direct coupling with DrawerFeature class.

src/main/java/edu/kis/powp/jobs2d/drivers/adapter/LineDrawerAdapter.java
```diff
@@ -0,0 +1,37 @@
+package edu.kis.powp.jobs2d.drivers.adapter;
+
+import edu.kis.legacy.drawer.panel.DrawPanelController;
+import edu.kis.legacy.drawer.shape.ILine;
+import edu.kis.powp.jobs2d.Job2dDriver;
+
+public class LineDrawerAdapter implements Job2dDriver {
+    private int startX = 0, startY = 0;
+    private DrawPanelController DrawingAdapter;
```
kowallus, line 9: fields names (in java convention) start from a lowercase

src/main/java/edu/kis/powp/jobs2d/drivers/adapter/DrawingAdapterDriver.java
```diff
@@ -4,14 +4,16 @@
...
 import edu.kis.legacy.drawer.shape.LineFactory;
 import edu.kis.powp.jobs2d.Job2dDriver;
+import edu.kis.powp.jobs2d.features.DrawerFeature;
+
 
 /**
  * driver adapter to drawer with several bugs.
  */
-public class MyAdapter extends DrawPanelController implements Job2dDriver {
+public class DrawingAdapterDriver extends DrawPanelController implements Job2dDriver {
```
kowallus, line 13: Inheritance is not used (no super drawLine calls). Should be removed!

kowallus (commented):
4.5 pts

PR #67, opened 2020-04-01
Tasks to 3.3 included

Graleszczur:


src/main/java/edu/kis/powp/jobs2d/drivers/adapter/LineDrawerAdapter.java
```diff
@@ -0,0 +1,53 @@
...
+    private int startX = 0;
+    private int startY = 0;
+
+    public LineDrawerAdapter(DrawPanelController drawPanelController, LineTypeEnum lineTypeEnum) {
+
+        this.drawPanelController = drawPanelController;
+
+        switch (lineTypeEnum) {
+            case BASIC:
+                this.line = LineFactory.getBasicLine();
```
kowallus, line 21: Not bad, but, this is not a generic solution. (e.g. using other line types is not yet possible)
I understand the idea, but such direct coupling is unnecessary (KISS principle)

src/main/java/edu/kis/powp/jobs2d/drivers/adapter/LineDrawerDriver.java
```diff
@@ -0,0 +1,38 @@
...
+
+import edu.kis.legacy.drawer.panel.DrawPanelController;
+import edu.kis.legacy.drawer.shape.ILine;
+import edu.kis.legacy.drawer.shape.LineFactory;
+import edu.kis.powp.jobs2d.Job2dDriver;
+
+/**
+ * driver adapter to drawer with several bugs.
+ */
+public class LineDrawerDriver extends DrawPanelController implements Job2dDriver {
```
kowallus, line 11: Inheritance is not used (no super drawLine calls). Should be removed!

kowallus (commented):

BTW: Whitespace modifications are really annoying.

PR #68, opened 2020-04-01
Lab1 - 3.3

ZB63:
Sebastian Pawełoszek
215844

### DIFF
--- a/src/main/java/edu/kis/powp/jobs2d/drivers/adapter/AdaptiveDrawAdapter.java
+++ b/src/main/java/edu/kis/powp/jobs2d/drivers/adapter/AdaptiveDrawAdapter.java
@@ -1,0 +1,40 @@
+package edu.kis.powp.jobs2d.drivers.adapter;
+
+import edu.kis.legacy.drawer.panel.DrawPanelController;
+import edu.kis.legacy.drawer.shape.ILine;
+import edu.kis.legacy.drawer.shape.LineFactory;
+import edu.kis.powp.jobs2d.Job2dDriver;
+import edu.kis.powp.jobs2d.features.DrawerFeature;
+
+/**
+ * driver adapter to drawer with several bugs.
+ */
+public class AdaptiveDrawAdapter extends DrawPanelController implements Job2dDriver {
+    private int startX = 0, startY = 0;
+    private ILine line;
+    public AdaptiveDrawAdapter(ILine line) {
+        super();
+        this.line = line;
+    }
+
+    @Override
+    public void setPosition(int x, int y) {
+        this.startX = x;
+        this.startY = y;
+    }
+
+    @Override
+    public void operateTo(int x, int y) {
+        line.setStartCoordinates(this.startX, this.startY);
+        line.setEndCoordinates(x, y);
+        setPosition(x, y);
+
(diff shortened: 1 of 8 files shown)


src/main/java/edu/kis/powp/jobs2d/drivers/adapter/AdaptiveDrawAdapter.java
```diff
@@ -0,0 +1,42 @@
...
+import edu.kis.legacy.drawer.panel.DrawPanelController;
+import edu.kis.legacy.drawer.shape.ILine;
+import edu.kis.legacy.drawer.shape.LineFactory;
+import edu.kis.powp.jobs2d.Job2dDriver;
+import edu.kis.powp.jobs2d.features.DrawerFeature;
+
+/**
+ * driver adapter to drawer with several bugs.
+ */
+public class AdaptiveDrawAdapter implements Job2dDriver {
```
kowallus, line 12: The name is better but it lacks a driver aspect.

src/main/java/edu/kis/powp/jobs2d/drivers/adapter/DrawAdapter.java
```diff
@@ -4,14 +4,15 @@
 import edu.kis.legacy.drawer.shape.ILine;
 import edu.kis.legacy.drawer.shape.LineFactory;
 import edu.kis.powp.jobs2d.Job2dDriver;
+import edu.kis.powp.jobs2d.features.DrawerFeature;
 
 /**
  * driver adapter to drawer with several bugs.
  */
-public class MyAdapter extends DrawPanelController implements Job2dDriver {
+public class DrawAdapter extends DrawPanelController implements Job2dDriver {
```
kowallus, line 12: The name is better but it lacks a driver aspect.

src/main/java/edu/kis/powp/jobs2d/drivers/adapter/DrawAdapter.java
```diff
@@ -26,8 +27,10 @@ public void operateTo(int x, int y) {
 		ILine line = LineFactory.getBasicLine();
 		line.setStartCoordinates(this.startX, this.startY);
 		line.setEndCoordinates(x, y);
+		setPosition(x, y);
 
-		drawLine(line);
+		DrawPanelController d = DrawerFeature.getDrawerController();
```
kowallus, line 32: Constructor injection should be used (using a parameter). This is unnecessary direct coupling with DrawerFeature class.

src/test/java/edu/kis/powp/jobs2d/events/SelectTestFigureOptionListener.java
```diff
@@ -5,7 +5,7 @@
 
 import edu.kis.powp.jobs2d.drivers.DriverManager;
 import edu.kis.powp.jobs2d.magicpresets.FiguresJoe;
-
+//Klasa jest aktualnie nieuzywana
```
kowallus, line 8: In revision repo it is safe to delete code instead of using a comment.

kowallus (commented):
3.5 pts

Incorrect relations on UML (inheritance instead of aggregation)

PR #69, opened 2020-04-01
Lab3.3

mateusz-git:


src/main/java/edu/kis/powp/jobs2d/drivers/adapter/DrawController.java
```diff
@@ -4,17 +4,18 @@
 import edu.kis.legacy.drawer.shape.ILine;
 import edu.kis.legacy.drawer.shape.LineFactory;
 import edu.kis.powp.jobs2d.Job2dDriver;
+import edu.kis.powp.jobs2d.features.DrawerFeature;
 
 /**
  * driver adapter to drawer with several bugs.
  */
-public class MyAdapter extends DrawPanelController implements Job2dDriver {
+public class DrawController extends DrawPanelController implements Job2dDriver {
```
kowallus, line 12: The name is better but it lacks a driver aspect.

src/main/java/edu/kis/powp/jobs2d/drivers/adapter/DrawController.java
```diff
@@ -4,17 +4,18 @@
 import edu.kis.legacy.drawer.shape.ILine;
 import edu.kis.legacy.drawer.shape.LineFactory;
 import edu.kis.powp.jobs2d.Job2dDriver;
+import edu.kis.powp.jobs2d.features.DrawerFeature;
```
kowallus, line 7: unused import

kowallus (commented):
4.5 pts

UML diagram is missing

PR #70, opened 2020-04-01
3.3

Patryk-97:
Zadanie 3.3 ukończone

### DIFF
--- a/src/main/java/edu/kis/powp/jobs2d/drivers/adapter/BasicLineDrawerAdapter.java
+++ b/src/main/java/edu/kis/powp/jobs2d/drivers/adapter/BasicLineDrawerAdapter.java
@@ -9,11 +9,11 @@ import edu.kis.powp.jobs2d.features.DrawerFeature;
 /**
  * driver adapter to drawer with several bugs.
  */
-public class MyAdapter implements Job2dDriver {
+public class BasicLineDrawerAdapter implements Job2dDriver {
 	private DrawPanelController drawPanelController;
 	private int startX = 0, startY = 0;
 
-	public MyAdapter() {
+	public BasicLineDrawerAdapter() {
 		super();
 		drawPanelController = DrawerFeature.getDrawerController();
 	}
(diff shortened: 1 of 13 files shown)


src/main/java/edu/kis/powp/jobs2d/drivers/adapter/SpecialLineDrawerAdapter.java
```diff
@@ -0,0 +1,45 @@
+package edu.kis.powp.jobs2d.drivers.adapter;
+
+import edu.kis.legacy.drawer.panel.DrawPanelController;
+import edu.kis.legacy.drawer.shape.ILine;
+import edu.kis.powp.jobs2d.Job2dDriver;
+
+public class SpecialLineDrawerAdapter implements Job2dDriver{
+	private DrawPanelController drawPanelController;
+	private int startX = 0, startY = 0;
+	private static ILine line = null;
```
kowallus, line 10: Why static?!

src/main/java/edu/kis/powp/jobs2d/drivers/adapter/BasicLineDrawerAdapter.java
```diff
@@ -8,11 +8,13 @@
 /**
  * driver adapter to drawer with several bugs.
  */
-public class MyAdapter extends DrawPanelController implements Job2dDriver {
+public class BasicLineDrawerAdapter implements Job2dDriver {
```
kowallus, line 11: The name is better but it lacks a driver aspect.

src/test/java/edu/kis/powp/jobs2d/events/SelectTestFigureOptionListener.java
```diff
@@ -16,6 +16,18 @@ public SelectTestFigureOptionListener(DriverManager driverManager) {
 
 	@Override
 	public void actionPerformed(ActionEvent e) {
-		FiguresJoe.figureScript1(driverManager.getCurrentDriver());
+		switch(e.getActionCommand())
+		{
+			case "Figure Joe 1":
```

kowallus (commented):
4.5 pts

PR #71, opened 2020-04-02
3.2, 3.3, 3.4*

ashleve:


ashleve:
> Direct coupling with DrawerFeature class is unnecessary.

> Hardcoded line type. What about 3.3.2 test?
> Unfortunately, this is not a generic solution. (e.g. new line type requires new class, unnecessary direct coupling)

I fixed the issues.

src/main/java/edu/kis/powp/jobs2d/drivers/adapter/LineDrawerAdapter.java
```diff
@@ -0,0 +1,52 @@
...
+    private DrawPanelController controller;
+    private ILine line;
+    private int startX = 0, startY = 0;
+
+    public LineDrawerAdapter(DrawPanelController controller, LineType lineType) {
+        this.controller = controller;
+
+        switch(lineType){
+            case BASIC:
+                this.line = LineFactory.getBasicLine();
```
kowallus, line 20: Coupling with LineFactory and LineType not necessary. KISS principle.

src/main/java/edu/kis/powp/jobs2d/drivers/adapter/MyAdapter.java
```diff
@@ -26,8 +27,9 @@ public void operateTo(int x, int y) {
 		ILine line = LineFactory.getBasicLine();
 		line.setStartCoordinates(this.startX, this.startY);
 		line.setEndCoordinates(x, y);
+		setPosition(x, y);
 
-		drawLine(line);
+		getDrawerController().drawLine(line);
```
kowallus, line 32: Constructor injection should be used (using a parameter). This is unnecessary direct coupling with DrawerFeature class.

src/main/java/edu/kis/powp/jobs2d/drivers/adapter/MyAdapter.java
```diff
@@ -4,11 +4,12 @@
 import edu.kis.legacy.drawer.shape.ILine;
 import edu.kis.legacy.drawer.shape.LineFactory;
 import edu.kis.powp.jobs2d.Job2dDriver;
+import edu.kis.powp.jobs2d.features.DrawerFeature;
 
 /**
  * driver adapter to drawer with several bugs.
  */
-public class MyAdapter extends DrawPanelController implements Job2dDriver {
+public class MyAdapter extends DrawerFeature implements Job2dDriver {
```
kowallus, line 12: Name not fixed

src/test/java/edu/kis/powp/jobs2d/events/SelectTestFigureOptionListener.java
```diff
@@ -16,6 +17,14 @@ public SelectTestFigureOptionListener(DriverManager driverManager) {
 
 	@Override
 	public void actionPerformed(ActionEvent e) {
-		FiguresJoe.figureScript1(driverManager.getCurrentDriver());
+		switch (e.getActionCommand()){
+			case "Figure Joe 1":
```

kowallus (commented):
4 pts

UML diagram is missing

PR #72, opened 2020-04-07
lab3 command and composite

Hergoln:


src/main/java/edu/kis/powp/command/ComplexCommand.java
```diff
@@ -0,0 +1,44 @@
...
+            command.execute();
+        }
+    }
+
+    public static Builder builder() {
+        return new Builder();
+    }
+
+    public static final class Builder {
+        private ArrayList<IDriverCommand> queue;
```
rsusik, line 27: List type (attribute and in the constructor) would be sufficient and would not limit the implementation.

rsusik (commented):

PR #74, opened 2020-04-07
Command & Composite

3elk4:


src/main/java/edu/kis/powp/command/ComplexCommand.java
```diff
@@ -0,0 +1,19 @@
+package edu.kis.powp.command;
+
+import edu.kis.powp.jobs2d.Job2dDriver;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class ComplexCommand implements DriverCommand {
+    List<DriverCommand> drivers = new ArrayList<>();
```
rsusik: Should be private

rsusik (commented):

PR #79, opened 2020-04-07
AdapterCheckout

Blackwatch98:
Zrobione zadania bez gwiazdek do 3.3.2


.idea/codeStyles/codeStyleConfig.xml
```diff
@@ -0,0 +1,5 @@
+<component name="ProjectCodeStyleConfiguration">
```

src/main/java/edu/kis/powp/jobs2d/drivers/adapter/DrawControllerAdapter.java
```diff
@@ -1,37 +1,43 @@
...
 import edu.kis.powp.jobs2d.Job2dDriver;
+import edu.kis.powp.jobs2d.features.DrawerFeature;
 
 /**
- * driver adapter to drawer with several bugs.
+ * Driver adapter to drawer with several bugs.
  */
-public class MyAdapter extends DrawPanelController implements Job2dDriver {
+
+public class DrawControllerAdapter extends DrawerFeature implements Job2dDriver
```
kowallus, line 11: Constructor injection should be used (using a parameter). This is unnecessary direct coupling with DrawerFeature class.

src/main/java/edu/kis/powp/jobs2d/drivers/adapter/LineDrawerAdapter.java
```diff
@@ -0,0 +1,60 @@
...
+    private int y = 0;
+
+    public LineDrawerAdapter(DrawPanelController controller, LineType lineType)
+    {
+        this.mainController = controller;
+
+        switch(lineType)
+        {
+            case SOLID:
+                this.lineToDraw = LineFactory.getBasicLine();
```
kowallus, line 25: Coupling with LineFactory and LineType. It is not necessary (KISS principle).

src/test/java/edu/kis/powp/jobs2d/TestJobs2dPatterns.java
```diff
@@ -83,7 +89,7 @@ public static void main(String[] args) {
 			public void run() {
 				Application app = new Application("2d jobs Visio");
 				DrawerFeature.setupDrawerPlugin(app);
-				setupDefaultDrawerVisibilityManagement(app);
+				//setupDefaultDrawerVisibilityManagement(app);
```
kowallus, line 92: In revision repo it is safe to delete code instead of using a comment.

src/test/java/edu/kis/powp/jobs2d/events/SelectTestFigureOptionListener.java
```diff
@@ -16,6 +17,13 @@ public SelectTestFigureOptionListener(DriverManager driverManager) {
 
 	@Override
 	public void actionPerformed(ActionEvent e) {
-		FiguresJoe.figureScript1(driverManager.getCurrentDriver());
+		if(e.paramString().contains("Figure Joe 2"))
```

kowallus (commented):
4 pts

PR #81, opened 2020-04-10
3.2 - 3.4.2

medykization:


.idea/codeStyles/codeStyleConfig.xml
```diff
@@ -0,0 +1,5 @@
+<component name="ProjectCodeStyleConfiguration">
```

src/main/java/edu/kis/powp/jobs2d/drivers/adapter/DrawerAdapter.java
```diff
@@ -4,15 +4,18 @@
...
  */
-public class MyAdapter extends DrawPanelController implements Job2dDriver {
+public class DrawerAdapter extends DrawPanelController implements Job2dDriver {
 	private int startX = 0, startY = 0;
+	private DrawPanelController drawController;
 
-	public MyAdapter() {
+	public DrawerAdapter() {
 		super();
+		drawController = DrawerFeature.getDrawerController();
```
kowallus, line 18: Constructor injection should be used (using a parameter). This is unnecessary direct coupling with DrawerFeature class.

src/main/java/edu/kis/powp/jobs2d/drivers/adapter/DrawerAdapter.java
```diff
@@ -4,15 +4,18 @@
 import edu.kis.legacy.drawer.shape.ILine;
 import edu.kis.legacy.drawer.shape.LineFactory;
 import edu.kis.powp.jobs2d.Job2dDriver;
+import edu.kis.powp.jobs2d.features.DrawerFeature;
 
 /**
  * driver adapter to drawer with several bugs.
  */
-public class MyAdapter extends DrawPanelController implements Job2dDriver {
+public class DrawerAdapter extends DrawPanelController implements Job2dDriver {
```
kowallus, line 12: Inheritance is not used (no super drawLine calls). Should be removed!

src/main/java/edu/kis/powp/jobs2d/drivers/adapter/LineDrawerAdapter.java
```diff
@@ -0,0 +1,36 @@
...
+        setPosition(x,y);
+
+        drawController.drawLine(this.iLine);
+    }
+
+    @Override
+    public String toString() {
+        return "special Line";
+    }
+}
```
kowallus, line 36: similar issues

src/test/java/edu/kis/powp/jobs2d/TestJobs2dPatterns.java
```diff
@@ -83,8 +92,7 @@ public static void main(String[] args) {
 			public void run() {
 				Application app = new Application("2d jobs Visio");
 				DrawerFeature.setupDrawerPlugin(app);
-				setupDefaultDrawerVisibilityManagement(app);
-
+				//setupDefaultDrawerVisibilityManagement(app); //
```
kowallus, line 95: In revision repo it is safe to delete code instead of using a comment.

src/test/java/edu/kis/powp/jobs2d/events/SelectTestFigureOptionListener.java
```diff
@@ -16,6 +18,22 @@ public SelectTestFigureOptionListener(DriverManager driverManager) {
 
 	@Override
 	public void actionPerformed(ActionEvent e) {
-		FiguresJoe.figureScript1(driverManager.getCurrentDriver());
+
+		String option = e.getActionCommand();
+		switch (option) {
+			case "Figure Joe 1":
```
kowallus, line 24: Menu action should not depend on its caption. Enum or separate listeners would be better. At least you should consider constants.

kowallus (commented):
3.5 pts

Incorrect composition on UML diagram.

PR #82, opened 2020-04-10
Lab adapter updated

MichalOgloza:
Michał Ogłoza 215834 grupa IO1


src/main/java/edu/kis/powp/jobs2d/drivers/adapter/LineDrawerAdapter.java
```diff
@@ -10,16 +10,13 @@
 /**
  * driver adapter to drawer with several bugs.
  */
-public class LineDrawerAdapter extends DrawPanelController implements Job2dDriver {
+public class LineDrawerAdapter implements Job2dDriver {
 	private int startX = 0, startY = 0;
 	private IModifiableLine line;
+	private DrawPanelController drawPanelController;
```
kowallus, line 16: You forgot to remove DrawerFeature import statement

kowallus (commented):
5 pts final

PR #84, opened 2020-04-11
3.2 3.3

TheDamik34:


src/main/java/edu/kis/powp/jobs2d/drivers/adapter/JobDriverToDrawer.java
```diff
@@ -4,14 +4,15 @@
 import edu.kis.legacy.drawer.shape.ILine;
 import edu.kis.legacy.drawer.shape.LineFactory;
 import edu.kis.powp.jobs2d.Job2dDriver;
+import edu.kis.powp.jobs2d.features.DrawerFeature;
 
 /**
  * driver adapter to drawer with several bugs.
  */
-public class MyAdapter extends DrawPanelController implements Job2dDriver {
+public class JobDriverToDrawer extends DrawPanelController implements Job2dDriver {
```
kowallus, line 12: Inheritance is not used (no super drawLine calls). Should be removed!

The name is hard to understand without mentioning "adapter" pattern

src/main/java/edu/kis/powp/jobs2d/drivers/adapter/JobDriverToDrawer.java
```diff
@@ -27,7 +28,9 @@ public void operateTo(int x, int y) {
 		line.setStartCoordinates(this.startX, this.startY);
 		line.setEndCoordinates(x, y);
 
-		drawLine(line);
+		this.startX = x;
+		this.startY = y;
```
kowallus, line 32: minor code redundancy

src/main/java/edu/kis/powp/jobs2d/drivers/adapter/JobDriverToDrawer.java
```diff
@@ -27,7 +28,9 @@ public void operateTo(int x, int y) {
...
+		DrawerFeature.getDrawerController().drawLine(line);
```
kowallus, line 33: Constructor injection should be used (using a parameter). This is unnecessary direct coupling with DrawerFeature class.

src/main/java/edu/kis/powp/jobs2d/drivers/adapter/LineDrawerAdapter.java
```diff
@@ -0,0 +1,38 @@
+package edu.kis.powp.jobs2d.drivers.adapter;
+
+import edu.kis.legacy.drawer.panel.DrawPanelController;
+import edu.kis.legacy.drawer.shape.ILine;
+import edu.kis.powp.jobs2d.Job2dDriver;
+
+public class LineDrawerAdapter extends DrawPanelController implements Job2dDriver {
```
kowallus, line 7: Inheritance is not used (no super drawLine calls). Should be removed!

src/test/java/edu/kis/powp/jobs2d/TestJobs2dPatterns.java
```diff
@@ -83,7 +91,7 @@ public static void main(String[] args) {
 			public void run() {
 				Application app = new Application("2d jobs Visio");
 				DrawerFeature.setupDrawerPlugin(app);
-				setupDefaultDrawerVisibilityManagement(app);
+				//setupDefaultDrawerVisibilityManagement(app);
```
kowallus, line 94: In revision repo it is safe to delete code instead of using a comment.

kowallus (commented):
2 pts

too late (2 pts penalty).

PR #89, opened 2020-04-15
Rozwiązanie instrukcji adapter

bopke:


src/main/java/edu/kis/powp/jobs2d/drivers/DriverManager.java
```diff
@@ -9,19 +9,19 @@
...
-	private Job2dDriver currentDriver = new LoggerDriver();
+    private Job2dDriver currentDriver = new LoggerDriver();
 
-	/**
-	 * @param driver Set the driver as current.
-	 */
-	public synchronized void setCurrentDriver(Job2dDriver driver) {
-		currentDriver = driver;
-	}
+    /**
```
kowallus, line 14: unnecessary format changes

src/main/java/edu/kis/powp/jobs2d/drivers/adapter/LineDrawerAdapter.java
```diff
@@ -0,0 +1,33 @@
+package edu.kis.powp.jobs2d.drivers.adapter;
+
+import edu.kis.legacy.drawer.shape.ILine;
+import edu.kis.powp.jobs2d.Job2dDriver;
+import edu.kis.powp.jobs2d.features.DrawerFeature;
+
+public class LineDrawerAdapter implements Job2dDriver {
+    private int startX = 0, startY = 0;
+    private ILine iLine;
```
kowallus, line 9: Originally, 'I' prefix has been meant to suggest an interface. This is a field so I would rather recommend skipping the first letter.

src/main/java/edu/kis/powp/jobs2d/drivers/adapter/LineDrawerAdapter.java
```diff
@@ -0,0 +1,33 @@
...
+        this.startX = x;
+        this.startY = y;
+    }
+
+    @Override
+    public void operateTo(int x, int y) {
+        iLine.setStartCoordinates(this.startX, this.startY);
+        iLine.setEndCoordinates(x, y);
+        setPosition(x, y);
+        DrawerFeature.getDrawerController().drawLine(iLine);
```
kowallus, line 26: Constructor injection should be used (using a parameter). This is unnecessary direct coupling with DrawerFeature class.

src/main/java/edu/kis/powp/jobs2d/drivers/adapter/ScreenDrawerDriver.java
```diff
@@ -0,0 +1,38 @@
...
+import edu.kis.legacy.drawer.panel.DrawPanelController;
+import edu.kis.legacy.drawer.shape.ILine;
+import edu.kis.legacy.drawer.shape.LineFactory;
+import edu.kis.powp.jobs2d.Job2dDriver;
+import edu.kis.powp.jobs2d.features.DrawerFeature;
+
+/**
+ * driver adapter to drawer with several bugs.
+ */
+public class ScreenDrawerDriver extends DrawPanelController implements Job2dDriver {
```
kowallus, line 12: Inheritance is not used (no super drawLine calls). Should be removed!

src/test/java/edu/kis/powp/jobs2d/TestJobs2dPatterns.java
```diff
@@ -7,92 +7,102 @@
...
+
+    /**
+     * Launch the application.
+     */
+    public static void main(String[] args) {
+        EventQueue.invokeLater(new Runnable() {
+            public void run() {
+                Application app = new Application("2d jobs Visio");
+                DrawerFeature.setupDrawerPlugin(app);
+                //setupDefaultDrawerVisibilityManagement(app);
```
kowallus, line 97: In revision repo it is safe to delete code instead of using a comment.

src/test/java/edu/kis/powp/jobs2d/events/SelectTestFigureOptionListener.java
```diff
@@ -4,18 +4,26 @@
...
+        this.driverManager = driverManager;
+    }
 
-	@Override
-	public void actionPerformed(ActionEvent e) {
-		FiguresJoe.figureScript1(driverManager.getCurrentDriver());
-	}
+    @Override
+    public void actionPerformed(ActionEvent e) {
+        if (e.getActionCommand().equals("Figure Joe 1")) {
```

kowallus (commented):
4 pts

UML should concern ScreenDrawerDriver.

PR #90, opened 2020-04-15
lab_command.

MichalOgloza:


kowallus:
5.5 pts

Ad. 3.1.6. I have expected a solution which would be generic (auto-genaration of DriverCommand from any code using Jobs2dDriver interface).

PR #91, opened 2020-04-15
Zadanie 2 - 3.1.5

maybegamer:
215693


src/main/java/edu/kis/powp/jobs2d/drivers/adapter/ComplexCommand.java
```diff
@@ -0,0 +1,20 @@
+package edu.kis.powp.jobs2d.drivers.adapter;
+
+import edu.kis.powp.jobs2d.Job2dDriver;
+
+import java.util.ArrayList;
+
+public class ComplexCommand implements DriverCommand {
+
+    ArrayList<DriverCommand> cascadeOfCommands = new ArrayList<>();
```
kowallus, line 9: List type (attribute and in the constructor) would be sufficient and would not limit the implementation.
Should be private

kowallus (commented):
4.5 pts

PR #92, opened 2020-04-16
Lab2 POWP

BartoszKowalczyk98:


src/main/java/edu/kis/powp/jobs2d/drivers/command/CommandBuilder.java
```diff
@@ -0,0 +1,19 @@
+package edu.kis.powp.jobs2d.drivers.command;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class CommandBuilder{
+    List<DriverCommand> commands;
```
rsusik, line 7: should be private

src/main/java/edu/kis/powp/jobs2d/drivers/command/ComplexCommand.java
```diff
@@ -0,0 +1,18 @@
+package edu.kis.powp.jobs2d.drivers.command;
+
+import java.util.List;
+
+public class ComplexCommand implements DriverCommand {
+    List<DriverCommand> driverCommandList;
```
rsusik, line 6: should be private

rsusik (commented):

PR #94, opened 2020-04-16
Command

Lisek8:


src/main/java/edu/kis/powp/jobs2d/command/ComplexCommand.java
```diff
@@ -0,0 +1,17 @@
+package edu.kis.powp.jobs2d.command;
+
+import java.util.List;
+
+public class ComplexCommand implements DriverCommand {
+    List<DriverCommand> commandList;
```
rsusik, line 6: should be private

src/main/java/edu/kis/powp/jobs2d/command/ComplexCommandBuilder.java
```diff
@@ -0,0 +1,23 @@
+package edu.kis.powp.jobs2d.command;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class ComplexCommandBuilder {
+    List<DriverCommand> commandList;
```
rsusik, line 7: should be private

rsusik (commented):

PR #96, opened 2020-04-18
Command

rysiuziom1:


src/main/java/edu/kis/powp/command/ComplexCommand.java
```diff
@@ -0,0 +1,16 @@
+package edu.kis.powp.command;
+
+import java.util.List;
+
+public class ComplexCommand implements DriverCommand{
+    List<DriverCommand> commands;
```
rsusik: Should be private

rysiuziom1: fixed in next commit

PR #97, opened 2020-04-19
Lab3 Command-Finished

Paulina-Koska:


src/main/java/edu/kis/powp/command/ComplexCommand.java
```diff
@@ -0,0 +1,17 @@
+package edu.kis.powp.command;
+
+import java.util.ArrayList;
+
+
+public class ComplexCommand implements DriverCommand {
+    ArrayList<DriverCommand> driverCommandList;
```
rsusik, line 7: should be private, List type (attribute and in the constructor) would be sufficient and would not limit the implementation.

rsusik (commented):

PR #98, opened 2020-04-19
Command Finished

wojciechciejka:


src/main/java/edu/kis/powp/command/ComplexCommand.java
```diff
@@ -0,0 +1,16 @@
+package edu.kis.powp.command;
+
+import java.util.List;
+
+public class ComplexCommand implements DriverCommand{
+    List<DriverCommand> driverCommandList;
```
rsusik, line 6: Should be private

pom.xml
```diff
@@ -1,49 +1,50 @@
-<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
```
rsusik, line 1: Commits polluted with project files

rsusik (commented):

PR #99, opened 2020-04-19
Lab3 Command.

Krizuu:
Command finished


.idea/compiler.xml
```diff
@@ -0,0 +1,16 @@
...
+        <sourceTestOutputDir name="target/generated-test-sources/test-annotations" />
+        <outputRelativeToContentRoot value="true" />
+        <module name="jobs2d" />
+      </profile>
+    </annotationProcessing>
+    <bytecodeTargetLevel>
+      <module name="jobs2d" target="1.8" />
+    </bytecodeTargetLevel>
+  </component>
+</project>
```
rsusik, line 16: Commits polluted with project files

src/main/java/edu/kis/powp/command/ComplexCommand.java
```diff
@@ -0,0 +1,44 @@
+package edu.kis.powp.command;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
+public class ComplexCommand implements DriverCommand {
+
+    private ArrayList<DriverCommand> queue;
```
rsusik, line 9: List type would be sufficient and would not limit the implementation.

rsusik (commented):

PR #100, opened 2020-04-20
Lab 3

Rorrim2:


src/main/java/edu/kis/powp/jobs2d/command/ComplexCommand.java
```diff
@@ -0,0 +1,16 @@
+package edu.kis.powp.jobs2d.command;
+
+import java.util.List;
+
+public class ComplexCommand implements DriverCommand{
+    List<DriverCommand> commands;
```
rsusik, line 6: should be private

rsusik (commented):

PR #102, opened 2020-04-20
Command/Composite

krzysiek1105:


.idea/codeStyles/codeStyleConfig.xml
```diff
@@ -0,0 +1,5 @@
+<component name="ProjectCodeStyleConfiguration">
+  <state>
+    <option name="PREFERRED_PROJECT_CODE_STYLE" value="Default" />
+  </state>
+</component>
```
rsusik, line 5: Commits polluted with project files

src/main/java/edu/kis/powp/command/CommandBuilder.java
```diff
@@ -0,0 +1,23 @@
+package edu.kis.powp.command;
+
+import java.util.ArrayList;
+
+public class CommandBuilder {
+    private ArrayList<DriverCommand> commands;
```
rsusik, line 6: List type would be sufficient and would not limit the implementation.

src/main/java/edu/kis/powp/command/ComplexCommandFactory.java
```diff
@@ -0,0 +1,36 @@
...
+        commandBuilder.add(new OperateToCommand(driver, x, y));
+
+        return commandBuilder.build();
+    }
+
+    public static DriverCommand circle(Job2dDriver driver, int x, int y, int r) {
+        CommandBuilder commandBuilder = new CommandBuilder();
+        double prevX = r;
+        double prevY = r;
+        commandBuilder.add(new SetPositionCommand(driver, (int) prevX + x, (int) prevY + y));
```
rsusik, line 21: r + x, r + y should be sufficient here

src/main/java/edu/kis/powp/command/ComplexCommand.java
```diff
@@ -0,0 +1,16 @@
+package edu.kis.powp.command;
+
+import java.util.ArrayList;
+
+public class ComplexCommand implements DriverCommand {
+    private ArrayList<DriverCommand> commands;
```
rsusik, line 6: List type would be sufficient and would not limit the implementation.

rsusik (commented):

PR #104, opened 2020-04-20
AdamWojtczak 215922 Command

AdamWojtczak:
AdamWojtczak 215922 Command

### DIFF
--- a/src/main/java/edu/kis/powp/jobs2d/drivers/command/CommandListBuilder.java
+++ b/src/main/java/edu/kis/powp/jobs2d/drivers/command/CommandListBuilder.java
@@ -1,0 +1,24 @@
+package edu.kis.powp.jobs2d.drivers.command;
+
+import java.util.LinkedList;
+import java.util.List;
+
+public class CommandListBuilder
+{
+    List<DriverCommand> commandsList;
+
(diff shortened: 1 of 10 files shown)


src/main/java/edu/kis/powp/jobs2d/drivers/command/CommandListBuilder.java
```diff
@@ -0,0 +1,24 @@
+package edu.kis.powp.jobs2d.drivers.command;
+
+import java.util.LinkedList;
+import java.util.List;
+
+public class CommandListBuilder
+{
+    List<DriverCommand> commandsList;
```
rsusik, line 8: should be private

src/main/java/edu/kis/powp/jobs2d/drivers/command/ComplexCommand.java
```diff
@@ -0,0 +1,22 @@
+package edu.kis.powp.jobs2d.drivers.command;
+
+import java.util.List;
+
+public class ComplexCommand implements DriverCommand
+{
+    List<DriverCommand> commandList;
```
rsusik, line 7: should be private

rsusik (commented):

PR #106, opened 2020-04-21
Command

BSkura98:


src/main/java/edu/kis/powp/command/ComplexCommand.java
```diff
@@ -0,0 +1,23 @@
+package edu.kis.powp.command;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class ComplexCommand implements DriverCommand{
+    List<DriverCommand> commandList;
```
rsusik, line 7: should be private

rsusik (commented):

PR #111, opened 2020-04-22
Laboratorium COMMAND

RafalProoo:


src/main/java/edu/kis/powp/jobs2d/command/ComplexCommand.java
```diff
@@ -0,0 +1,19 @@
+package edu.kis.powp.jobs2d.command;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class ComplexCommand implements DriverCommand {
+
+    private List<DriverCommand> driverCommands;
+
+    public ComplexCommand(ArrayList<DriverCommand> driverCommands) {
```
kowallus, line 10: type of param should be List

src/main/java/edu/kis/powp/jobs2d/command/OperateToCommand.java
```diff
@@ -0,0 +1,20 @@
+package edu.kis.powp.jobs2d.command;
+
+import edu.kis.powp.jobs2d.Job2dDriver;
+
+public class OperateToCommand implements DriverCommand {
+
+    private int X;
```
kowallus, line 7: in java field names start from a lowercase (x)

src/main/java/edu/kis/powp/jobs2d/command/ShapeFactory.java
```diff
@@ -0,0 +1,44 @@
+package edu.kis.powp.jobs2d.command;
+
+import edu.kis.powp.jobs2d.Job2dDriver;
+
+import java.util.ArrayList;
+
+public class ShapeFactory {
+
+    public static DriverCommand makeRectangle(Job2dDriver job2dDriver) {
+        ArrayList<DriverCommand> driverCommandList = new ArrayList<>();
```
kowallus, line 10: Type should indicate interface, i.e. List.

kowallus (commented):
4.5 pts

PR #112, opened 2020-04-22
Command

GGuinea:
Lab 3.3


src/main/java/edu/kis/powp/jobs2d/command/ComplexCommand.java
```diff
@@ -0,0 +1,35 @@
+package edu.kis.powp.jobs2d.command;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class ComplexCommand implements DriverCommand {
+
+    private List<DriverCommand> commandList = new ArrayList<>();
+
+    public ComplexCommand(ArrayList<DriverCommand> commandList) {
```
kowallus, line 10: List type should be used!

kowallus (commented):
5.5 pts (bonus for a circle)

PR #113, opened 2020-04-22
Command, Composite

ZB63:
Sebastian Paweloszek 
215844

### DIFF
--- a/src/main/java/edu/kis/powp/command/ComplexCommand.java
+++ b/src/main/java/edu/kis/powp/command/ComplexCommand.java
@@ -4,10 +4,10 @@ import java.util.ArrayList;
 
 public class ComplexCommand implements DriverCommand {
 
-    ArrayList<DriverCommand> commandList;
+    private ArrayList<DriverCommand> commandList;
 
-    public ComplexCommand(ArrayList<DriverCommand> arrayList) {
-        this.commandList = arrayList;
+    public ComplexCommand(ArrayList<DriverCommand> commands) {
+        this.commandList = commands;
     }
 
     @Override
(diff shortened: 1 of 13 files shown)


src/main/java/edu/kis/powp/command/ComplexCommand.java
```diff
@@ -0,0 +1,19 @@
+package edu.kis.powp.command;
+
+import java.util.ArrayList;
+
+public class ComplexCommand implements DriverCommand {
+
+    private ArrayList<DriverCommand> commandList;
```
kowallus, line 7: List type (attribute and in the constructor) would be sufficient and would not limit the implementation.

src/main/java/edu/kis/powp/command/ComplexCommand.java
```diff
@@ -0,0 +1,19 @@
...
+
+    public ComplexCommand(ArrayList<DriverCommand> commands) {
```
kowallus, line 9: also List

src/main/java/edu/kis/powp/command/ComplexCommandsFactory.java
```diff
@@ -0,0 +1,54 @@
...
+import java.util.ArrayList;
+
+public class ComplexCommandsFactory {
+
+    private ComplexCommandsFactory(){
+
+    }
+
+    public static ComplexCommand createRectangle(Job2dDriver driver) {
+        ArrayList<DriverCommand> commands = new ArrayList<DriverCommand>();
```
kowallus, line 14: again List type would be better

kowallus (commented):
5- pts

PR #114, opened 2020-04-22
Command PR

Kacpri:


src/main/java/edu/kis/powp/jobs2d/drivers/command/ComplexCommand.java
```diff
@@ -0,0 +1,21 @@
+package edu.kis.powp.jobs2d.drivers.command;
+
+import edu.kis.powp.jobs2d.Job2dDriver;
+
+import java.util.LinkedList;
+import java.util.List;
+
+public class ComplexCommand implements DriverCommand {
+
+    private final LinkedList<DriverCommand> commands;
```
kowallus, line 10: List type (attribute and in the constructor) would be sufficient and would not limit the implementation.

kowallus (commented):
5 pts

PR #115, opened 2020-04-22
Lab adapter

RadWex:


.idea/compiler.xml
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
```

src/main/java/edu/kis/powp/jobs2d/drivers/adapter/DrawLineAdapter.java
```diff
@@ -8,11 +8,14 @@
 /**
  * driver adapter to drawer with several bugs.
  */
-public class MyAdapter extends DrawPanelController implements Job2dDriver {
+public class DrawLineAdapter extends DrawPanelController implements Job2dDriver {
```
kowallus, line 11: Inheritance is not used (no super drawLine calls). Should be removed!

src/main/java/edu/kis/powp/jobs2d/drivers/adapter/LineDrawerAdapter.java
```diff
@@ -0,0 +1,34 @@
...
+    public void setPosition(int x, int y) {
+        this.startX = x;
+        this.startY = y;
+    }
+
+    @Override
+    public void operateTo(int x, int y) {
+        this.line.setStartCoordinates(this.startX, this.startY);
+        this.line.setEndCoordinates(x, y);
+        DrawerFeature.getDrawerController().drawLine(this.line);
```
kowallus, line 26: Constructor injection should be used (using a parameter). This is unnecessary direct coupling with DrawerFeature class.

src/test/java/edu/kis/powp/jobs2d/events/SelectTestFigureOptionListener.java
```diff
@@ -16,6 +16,11 @@ public SelectTestFigureOptionListener(DriverManager driverManager) {
 
 	@Override
 	public void actionPerformed(ActionEvent e) {
-		FiguresJoe.figureScript1(driverManager.getCurrentDriver());
+		if(e.getActionCommand().equals("Figure Joe 1")) {
```

src/test/java/edu/kis/powp/jobs2d/TestJobs2dPatterns.java
```diff
@@ -54,7 +63,7 @@ private static void setupDefaultDrawerVisibilityManagement(Application applicati
 		DefaultDrawerFrame defaultDrawerWindow = DefaultDrawerFrame.getDefaultDrawerFrame();
 		application.addComponentMenuElementWithCheckBox(DrawPanelController.class, "Default Drawer Visibility",
 				new SelectChangeVisibleOptionListener(defaultDrawerWindow), true);
-		defaultDrawerWindow.setVisible(true);
+		//defaultDrawerWindow.setVisible(true);
```
kowallus, line 66: In revision repo it is safe to delete code instead of using a comment.

kowallus (commented):
2.5 pts (too late -1 pts)

A small number of commits with confusing names.

PR #118, opened 2020-04-24
Lab2: command, composite

startled-cat:
Adam Kowalczyk
215771


src/main/java/edu/kis/powp/jobs2d/drivers/command/CommandFactory.java
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    public static DriverCommand getDiamondCommand(Job2dDriver job2dDriver){
+        ComplexCommandBuilder complexCommandBuilder = new ComplexCommandBuilder();
+
+        //middle
+        int x0 = 0;
+        int y0 = 0;
+        //size
+        int size = 200;
```
rsusik, line 57: Constants should be considered if it never changes.

src/main/java/edu/kis/powp/jobs2d/drivers/command/ComplexCommandBuilder.java
```diff
@@ -0,0 +1,23 @@
+package edu.kis.powp.jobs2d.drivers.command;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class ComplexCommandBuilder {
+    List<DriverCommand> commandList;
```
rsusik, line 7: should be private

rsusik (commented):

PR #120, opened 2020-04-24
Command

matlukomski94:


src/main/java/edu/kis/powp/jobs2d/command/ComplexCommand.java
```diff
@@ -0,0 +1,20 @@
+package edu.kis.powp.jobs2d.command;
+
+
+import java.util.ArrayList;
+
+public class ComplexCommand implements DriverCommand {
+
+    private ArrayList<DriverCommand> driverCommandList;
```
rsusik, line 8: List type would be sufficient and would not limit the implementation.

rsusik (commented):

PR #121, opened 2020-04-25
command

Mmati1996:
command

### DIFF
--- a/src/main/java/Commands/ComplexCommand.java
+++ b/src/main/java/Commands/ComplexCommand.java
@@ -1,0 +1,26 @@
+package Commands;
+
+import java.util.ArrayList;
+
+public class ComplexCommand {
+    ArrayList<DriverCommand> commands;
+
+    public ComplexCommand(ArrayList<DriverCommand> commands) {
+        this.commands = commands;
+    }
+
+    public ComplexCommand(){
+        this.commands = new ArrayList<DriverCommand>();
+    }
+
+    public void addCommand(DriverCommand command){
+        this.commands.add(command);
+    }
+
+    public void execute(){
+        for (int i = 0; i<commands.size();i++){
+            commands.get(i).execute();
+        }
+    }
+
+}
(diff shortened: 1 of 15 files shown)


src/main/java/Commands/ComplexCommand.java
```diff
@@ -0,0 +1,26 @@
+package Commands;
+
+import java.util.ArrayList;
+
+public class ComplexCommand {
+    ArrayList<DriverCommand> commands;
```
kowallus, line 6: Should be private.
List type (attribute and in the constructor) would be sufficient and would not limit the implementation.

src/main/java/Commands/OperateToCommand.java
```diff
@@ -0,0 +1,20 @@
+package Commands;
+
+import edu.kis.powp.jobs2d.Job2dDriver;
+
+public class OperateToCommand implements DriverCommand {
+    int x;
```
kowallus, line 6: All fields should be private

src/main/java/factories/TriangleFactory.java
```diff
@@ -0,0 +1,28 @@
+package factories;
+
+import Commands.ComplexCommand;
+import Commands.OperateToCommand;
+import Commands.SetPositionCommand;
+import edu.kis.powp.jobs2d.drivers.adapter.Pane2dDrawerAdapter;
+
+public class TriangleFactory {
```
kowallus, line 8: Not tested?!

src/main/java/factories/TriangleFactory.java
```diff
@@ -0,0 +1,28 @@
...
+
+    ComplexCommand triangle;
+
+    public ComplexCommand createTriangle(){
+        triangle = new ComplexCommand();
+        triangle.addCommand(new SetPositionCommand(0,0,new Pane2dDrawerAdapter()));
```
kowallus, line 14: each command has a different driver?!

kowallus (commented):
3 pts

PR #123, opened 2020-04-26
Command

Nyitz:


src/main/java/edu/kis/powp/jobs2d/command/ComplexCommand.java
```diff
@@ -0,0 +1,18 @@
+package edu.kis.powp.jobs2d.command;
+
+import java.util.ArrayList;
+
+public class ComplexCommand implements DriverCommand {
+    private ArrayList<DriverCommand> driverCommands;
```
rsusik, line 6: List type would be sufficient and would not limit the implementation.

src/test/java/edu/kis/powp/jobs2d/events/SelectTestFigureOptionListener.java
```diff
@@ -16,6 +16,9 @@ public SelectTestFigureOptionListener(DriverManager driverManager) {
 
 	@Override
 	public void actionPerformed(ActionEvent e) {
-		FiguresJoe.figureScript1(driverManager.getCurrentDriver());
+		if (e.getActionCommand().equals("Figure Joe 1"))
+			FiguresJoe.figureScript1(driverManager.getCurrentDriver());
+		if (e.getActionCommand().equals("Figure Joe 2"))
+			FiguresJoe.figureScript2(driverManager.getCurrentDriver());
```
rsusik, line 22: Still hardcoded. Enums or constants should be used.

rsusik (commented):

PR #125, opened 2020-04-26
Command

zenakz:


src/main/java/edu/kis/powp/jobs2d/command/ComplexBuilder.java
```diff
@@ -0,0 +1,17 @@
+package edu.kis.powp.jobs2d.command;
+
+import java.util.LinkedList;
+
+public class ComplexBuilder {
+    private LinkedList<DriverCommand> list=new LinkedList<>();
```
rsusik, line 6: List type would be sufficient and would not limit the implementation.

src/main/java/edu/kis/powp/jobs2d/command/ComplexBuilder.java
```diff
@@ -0,0 +1,17 @@
...
+
+    void addCommand(DriverCommand command)
+    {
+        list.add(command);
```
rsusik, line 10: It is recommended to return "this" so you can call methods in sequence.

src/main/java/edu/kis/powp/jobs2d/command/ComplexCommand.java
```diff
@@ -0,0 +1,17 @@
+package edu.kis.powp.jobs2d.command;
+import java.util.LinkedList;
+
+public class ComplexCommand implements DriverCommand {
+
+    private LinkedList<DriverCommand>list;
```
rsusik, line 6: List type would be sufficient and would not limit the implementation.

src/main/java/edu/kis/powp/jobs2d/command/OperateToCommand.java
```diff
@@ -0,0 +1,18 @@
+package edu.kis.powp.jobs2d.command;
+
+import edu.kis.powp.jobs2d.Job2dDriver;
+
+public class OperateToCommand implements DriverCommand {
+    int x;
+    int y;
+    Job2dDriver driver;
```
rsusik, line 8: should be private

src/main/java/edu/kis/powp/jobs2d/command/SetPositionCommand.java
```diff
@@ -0,0 +1,19 @@
+package edu.kis.powp.jobs2d.command;
+
+import edu.kis.powp.jobs2d.Job2dDriver;
+
+public class SetPositionCommand implements DriverCommand {
+    int x;
+    int y;
+    Job2dDriver driver;
```
rsusik, line 8: should be private

src/test/java/edu/kis/powp/jobs2d/events/SelectTestFigureOptionListener.java
```diff
@@ -16,6 +17,8 @@ public SelectTestFigureOptionListener(DriverManager driverManager) {
 
 	@Override
 	public void actionPerformed(ActionEvent e) {
-		FiguresJoe.figureScript1(driverManager.getCurrentDriver());
+		DrawerFeature.getDrawerController().clearPanel();
+		if(e.getActionCommand().equals("Figure Joe 2"))FiguresJoe.figureScript2(driverManager.getCurrentDriver());
```
rsusik, line 21: Still hardcoded. Should be enum or constant.

rsusik (commented):

PR #126, opened 2020-04-27
Lab Command

DeanMiles:
Alan Franas
215717


src/main/java/edu/kis/powp/jobs2d/commands/ComplexCommand.java
```diff
@@ -0,0 +1,28 @@
+package edu.kis.powp.jobs2d.commands;
+
+import java.util.ArrayList;
+
+public class ComplexCommand implements DriverCommand {
+
+    private ArrayList<DriverCommand> commands = new ArrayList<>();
```
kowallus, line 7: List type (attribute and in the constructor) would be sufficient and would not limit the implementation.

src/main/java/edu/kis/powp/jobs2d/commands/ComplexCommand.java
```diff
@@ -0,0 +1,28 @@
...
+
+    public ComplexCommand() {
+    }
+
+    public ComplexCommand(ArrayList<DriverCommand> commands) {
```
kowallus, line 12: also List

kowallus (commented):
4 pts (too late -1 pts)

DeanMiles:
> 4 pts (too late -1 pts)
@kowallus 
command

Hello, I do not undestand why I have lost point If the information on Wikamp says that I did it on time. 
Best regards,
Alan Franas

Edit. I also did not found information about other deadline in group chat at Teams.

PR #128, opened 2020-04-27
Command 3.1

Smirnoffq:


src/main/java/edu/kis/powp/jobs2d/drivers/command/ComplexCommand.java
```diff
@@ -0,0 +1,22 @@
+package edu.kis.powp.jobs2d.drivers.command;
+
+import java.util.List;
+
+public class ComplexCommand implements DriverCommand
+{
+    List<DriverCommand> commandList;
```
rsusik, line 7: should be private

src/main/java/edu/kis/powp/jobs2d/drivers/command/ListBuilder.java
```diff
@@ -0,0 +1,24 @@
+package edu.kis.powp.jobs2d.drivers.command;
+
+import java.util.LinkedList;
+import java.util.List;
+
+public class ListBuilder
+{
+    List<DriverCommand> commandsList;
```
rsusik, line 8: same here - should be private

src/main/java/edu/kis/powp/jobs2d/drivers/command/OperateToCommand.java
```diff
@@ -0,0 +1,22 @@
+package edu.kis.powp.jobs2d.drivers.command;
+
+import edu.kis.powp.jobs2d.Job2dDriver;
+
+public class OperateToCommand implements DriverCommand
+{
+    int x, y;
+    Job2dDriver job2dDriver;
```
rsusik, line 8: and the same here - should be private

src/test/java/edu/kis/powp/jobs2d/events/SelectTestFigureOptionListener.java
```diff
@@ -16,6 +16,13 @@ public SelectTestFigureOptionListener(DriverManager driverManager) {
 
 	@Override
 	public void actionPerformed(ActionEvent e) {
-		FiguresJoe.figureScript1(driverManager.getCurrentDriver());
+		switch(e.getActionCommand()) {
+			case "Figure Joe 1":
```
rsusik, line 20: This is still hardcoded, Enums or constants would be better for such purposes.

src/main/java/edu/kis/powp/jobs2d/drivers/command/SetPositionCommand.java
```diff
@@ -0,0 +1,22 @@
+package edu.kis.powp.jobs2d.drivers.command;
+
+import edu.kis.powp.jobs2d.Job2dDriver;
+
+public class SetPositionCommand implements DriverCommand
+{
+    int x, y;
+    Job2dDriver job2dDriver;
```
rsusik, line 8: should be private

rsusik (commented):
The solution sent late

PR #129, opened 2020-04-27
Command

Thomson19:


src/main/java/edu/kis/powp/jobs2d/drivers/command/ComplexCommand.java
```diff
@@ -0,0 +1,24 @@
+package edu.kis.powp.jobs2d.drivers.command;
+
+import edu.kis.powp.jobs2d.Job2dDriver;
+
+import java.util.ArrayList;
+
+public class ComplexCommand implements DriverCommand {
+    private Job2dDriver job2dDriver;
+    private ArrayList<DriverCommand> commandList;
```
rsusik, line 9: List type would be sufficient and would not limit the implementation.

src/test/java/edu/kis/powp/jobs2d/events/SelectTestFigureOptionListener.java
```diff
@@ -16,6 +18,19 @@ public SelectTestFigureOptionListener(DriverManager driverManager) {
...
+				break;
+			case "Figure Joe 2":
+				FiguresJoe.figureScript2(driverManager.getCurrentDriver());
+				break;
+			case "Square":
+				ComplexCommandFactory.Square(driverManager.getCurrentDriver(), 10, 10, 10).execute();
+				break;
+			case "Rectangle":
+				ComplexCommandFactory.Rectangle(driverManager.getCurrentDriver(), 50, 50, 100, 50).execute();
+				break;
```
rsusik, line 33: Still hardcoded. Enums or constants would be better for such purposes.

rsusik (commented):
The solution sent late

PR #131, opened 2020-04-27
command 3.1.5

pawellazicki:


src/main/java/edu/kis/powp/jobs2d/drivers/Command/ComplexCommand.java
```diff
@@ -0,0 +1,19 @@
+package edu.kis.powp.jobs2d.drivers.Command;
+
+import java.util.LinkedList;
+import java.util.List;
+
+public class ComplexCommand implements DriverCommand {
+    List<DriverCommand> commandList = new LinkedList<>();
```
kowallus, line 7: should be private

src/main/java/edu/kis/powp/jobs2d/figures/Figures.java
```diff
@@ -0,0 +1,28 @@
...
+    public static void figureScript2(Job2dDriver job2dDriver) {
+        ComplexCommand complexCommand = new ComplexCommand();
+
+        complexCommand.addCommand(new SetPositionCommand(0, 0, job2dDriver));
+        complexCommand.addCommand(new OperateToCommand(0, 50, job2dDriver));
+        complexCommand.addCommand(new OperateToCommand(50, 50, job2dDriver));
+        complexCommand.addCommand(new OperateToCommand(50, 0, job2dDriver));
+        complexCommand.addCommand(new OperateToCommand(0, 0, job2dDriver));
+
+        complexCommand.execute();
```
kowallus, line 26: a factory should return a complex command.

src/test/java/edu/kis/powp/jobs2d/events/SelectTestFigureOptionListener.java
```diff
@@ -18,8 +19,15 @@ public SelectTestFigureOptionListener(DriverManager driverManager) {
 	public void actionPerformed(ActionEvent e) {
 		if (e.getActionCommand().toLowerCase().equals("figure joe 1")) {
 			FiguresJoe.figureScript1(driverManager.getCurrentDriver());
-		} else if (e.getActionCommand().toLowerCase().equals("figure joe 2")) {
+		}
+		else if (e.getActionCommand().toLowerCase().equals("figure joe 2")) {
 			FiguresJoe.figureScript2(driverManager.getCurrentDriver());
 		}
+		else if(e.getActionCommand().toLowerCase().equals("figure command")){
```
kowallus, line 26: Menu action should not depend on its caption. Enum or separate listeners would be better. At least you should consider constants.

kowallus (commented):
3.5 pts

PR #132, opened 2020-04-27
Lab3 Command

Patryk-97:
Lab3 Command do 3.1.5 włącznie

### DIFF
--- a/src/main/java/edu/kis/powp/command/ComplexCommand.java
+++ b/src/main/java/edu/kis/powp/command/ComplexCommand.java
@@ -1,0 +1,41 @@
+package edu.kis.powp.command;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class ComplexCommand implements DriverCommand {
+	
+	private ComplexCommand(List<DriverCommand> commandsList) {
+		this.commandsList = commandsList;
+	}
+	
+	@Override
+	public void execute() {
+		for(DriverCommand driverCommand : commandsList) {
+			driverCommand.execute();
+		}
+	}
+	
+	public static Builder builder() {
+		return new Builder();
+	}
+	
+	public static final class Builder {
+		
+		private Builder() {
+			commandsList = new ArrayList<>();
+		}
+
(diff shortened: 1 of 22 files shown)


src/main/java/edu/kis/powp/command/ComplexCommand.java
```diff
@@ -0,0 +1,42 @@
...
+			commandsList = new ArrayList<>();
+		}
+
+		public Builder addCommand(DriverCommand driverCommand) {
+			commandsList.add(driverCommand);
+			return this;
+		}
+
+		public ComplexCommand build() {
+			return new ComplexCommand(commandsList);
```
kowallus, line 35: Nice but it would be best to make o copy of the original list.

src/main/java/edu/kis/powp/jobs2d/drivers/CircleFigure.java
```diff
@@ -0,0 +1,50 @@
...
+	public CircleFigure(Job2dDriver driver, int circleCenterX, int circleCenterY, int radius) {
+		this.builder = ComplexCommand.builder();
+		this.driver = driver;
+		this.circleCenterX = circleCenterX;
+		this.circleCenterY = circleCenterY;
+		this.radius = radius;
+	}
+	
+	@Override
+	public DriverCommand draw() {
```
kowallus, line 20: the name - draw - seems incorrect (e.g. getFigure)

src/main/java/edu/kis/powp/jobs2d/drivers/FiguresFactory.java
```diff
@@ -0,0 +1,29 @@
+package edu.kis.powp.jobs2d.drivers;
+
+import edu.kis.powp.jobs2d.Job2dDriver;
+
+import edu.kis.powp.command.DriverCommand;
+
+public class FiguresFactory {
```
kowallus, line 7: This class seems redundant.

kowallus (commented):
6 pts (bonus for factory method and circle)

PR #133, opened 2020-04-28
Lab Command

mati9844:


src/main/java/edu/kis/powp/jobs2d/command/ComplexCommand.java
```diff
@@ -0,0 +1,22 @@
+package edu.kis.powp.jobs2d.command;
+
+import java.util.Vector;
+
+public class ComplexCommand implements DriverCommand {
+    private Vector<DriverCommand> commandSeq = new Vector<DriverCommand>();
```
kowallus, line 6: List type (attribute and in the constructor) would be sufficient and would not limit the implementation.

src/test/java/edu/kis/powp/jobs2d/events/SelectTestFigureOptionListener.java
```diff
@@ -16,6 +20,26 @@ public SelectTestFigureOptionListener(DriverManager driverManager) {
 
 	@Override
 	public void actionPerformed(ActionEvent e) {
-		FiguresJoe.figureScript1(driverManager.getCurrentDriver());
+		if (e.getActionCommand().toLowerCase().equals("figure joe 1")) {
```
kowallus, line 23: Menu action should not depend on its caption. Enum or separate listeners would be better. At least you should consider constants.

kowallus (commented):
5.5 pts

Currently, there is no abstract factory in this project.

PR #134, opened 2020-04-28
Lab Command 3.1 - 3.7*

ashleve:


src/main/java/edu/kis/powp/jobs2d/builders/ComplexCommandBuilder.java
```diff
@@ -0,0 +1,27 @@
...
+        commands = new ArrayList<>();
+    }
+
+    public void addCommand(DriverCommand command){
+        commands.add(command);
+    }
+
+    @Override
+    public ComplexCommand build(){
+        return new ComplexCommand(commands);
```
kowallus, line 23: Nice, but it would be better to return a copy of commands.

kowallus (commented):
6 pts (bonus for factory method and observer)

Ad. 3.1.6. I have expected a solution which would be generic (auto-genaration of DriverCommand from any code using Jobs2dDriver interface).

PR #136, opened 2020-04-28
zad2

cyngiel:


.idea/compiler.xml
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
```

src/main/java/edu/kis/powp/jobs2d/drivers/Command/ComplexCommand.java
```diff
@@ -0,0 +1,26 @@
+package edu.kis.powp.jobs2d.drivers.Command;
+
+import edu.kis.powp.jobs2d.Job2dDriver;
+
+import java.util.ArrayList;
+
+public class ComplexCommand implements DriverCommand {
+
+    ArrayList<DriverCommand> commands;
```
kowallus, line 9: Should be private.
List type (attribute and in the constructor) would be sufficient and would not limit the implementation.

src/main/java/edu/kis/powp/jobs2d/drivers/Command/ComplexCommand.java
```diff
@@ -0,0 +1,26 @@
...
+        commands = new ArrayList<>();
+    }
+
+    public void addCommand(DriverCommand command){
+        commands.add(command);
+    }
+
+    @Override
+    public void execute(Job2dDriver job2dDriver) {
+        System.out.println("complex execute");
```
kowallus, line 21: should be removed (finally)

src/main/java/edu/kis/powp/jobs2d/drivers/Command/ComplexCommand.java
```diff
@@ -0,0 +1,26 @@
...
+        for(int i = 0; i < commands.size(); i++){
```
kowallus, line 22: foreach statement would be much simpler

src/main/java/edu/kis/powp/jobs2d/drivers/Command/OperateToCommand.java
```diff
@@ -0,0 +1,18 @@
+package edu.kis.powp.jobs2d.drivers.Command;
+
+import edu.kis.powp.jobs2d.Job2dDriver;
+
+public class OperateToCommand implements DriverCommand{
+
+    int x, y;
```
kowallus, line 7: should be private

src/main/java/edu/kis/powp/jobs2d/drivers/Command/SandglassFactory.java
```diff
@@ -0,0 +1,19 @@
+package edu.kis.powp.jobs2d.drivers.Command;
+
+import edu.kis.powp.jobs2d.Job2dDriver;
+
+public class SandglassFactory {
+
+    ComplexCommand command;
```
kowallus, line 7: should be local in run method!

src/test/java/edu/kis/powp/jobs2d/events/SelectTestFigureOptionListener.java
```diff
@@ -22,5 +28,13 @@ public void actionPerformed(ActionEvent e) {
 		} else if (e.getActionCommand().toLowerCase().equals("figure joe 2")) {
 			FiguresJoe.figureScript2(driverManager.getCurrentDriver());
 		}
+		else if (e.getActionCommand().toLowerCase().equals("rectangle factory")) {
```
kowallus, line 31: Menu action should not depend on its caption. Enum or separate listeners would be better. At least you should consider constants.

src/main/java/edu/kis/powp/jobs2d/drivers/Command/SandglassFactory.java
```diff
@@ -0,0 +1,19 @@
...
+
+    public void run(Job2dDriver job2dDriver){
```
kowallus, line 9: factories should return commands (not execute them).

kowallus (commented):
3 pts

PR #137, opened 2020-04-28
Lab2 Command design pattern

bopke:


src/main/java/edu/kis/powp/jobs2d/commands/OperateToCommand.java
```diff
@@ -0,0 +1,36 @@
...
+    @Override
+    public void execute() {
+        driver.operateTo(x, y);
+    }
+
+    public int getX() {
+        return x;
+    }
+
+    public OperateToCommand setX(int x) {
```
kowallus, line 23: I don't think that fluent syntax, in this case, is justified. Setting X and Y in the constructor is enough.

src/main/java/edu/kis/powp/jobs2d/drivers/SelectDriverMenuOptionListener.java
```diff
@@ -7,17 +7,17 @@
 import edu.kis.powp.jobs2d.features.DriverFeature;
 
 public class SelectDriverMenuOptionListener implements ActionListener {
-	private DriverManager driverManager;
-	private Job2dDriver driver = null;
+    private DriverManager driverManager;
```
kowallus, line 10: changing the convention of the original code is a bad habit.

src/main/java/edu/kis/powp/jobs2d/drivers/adapter/LineDrawerAdapter.java
```diff
@@ -0,0 +1,33 @@
+package edu.kis.powp.jobs2d.drivers.adapter;
+
+import edu.kis.legacy.drawer.shape.ILine;
+import edu.kis.powp.jobs2d.Job2dDriver;
+import edu.kis.powp.jobs2d.features.DrawerFeature;
+
+public class LineDrawerAdapter implements Job2dDriver {
+    private int startX = 0, startY = 0;
+    private ILine iLine;
```
kowallus, line 9: 'I' in the type name stands for 'interface'. It should not be applied to a field name ('line' would be better)

src/test/java/edu/kis/powp/jobs2d/events/SelectTestFigureOptionListener.java
```diff
@@ -24,6 +25,10 @@ public void actionPerformed(ActionEvent e) {
             FiguresJoe.figureScript2(driverManager.getCurrentDriver());
         } else if (e.getActionCommand().equals("Figure Jane 1")) {
             FiguresJane.figureScript(new DriverAdapter(driverManager.getCurrentDriver()));
+        } else if (e.getActionCommand().equals("Square")) {
```
kowallus, line 28: Menu action should not depend on its caption. Enum or separate listeners would be better. At least you should consider constants.

kowallus (commented):
3.5 pts (too late -1 pts)

PR #138, opened 2020-04-28
Lab 3 command 

tobiaszkubiak31:
Lab3 Command do 3.1.5, 3.1.6 nie jestem pewien
3.1.4 Composite design pattern


src/test/java/edu/kis/powp/jobs2d/TestJobs2dPatterns.java
```diff
@@ -19,99 +22,118 @@
...
+    private final static Logger logger = Logger.getLogger(Logger.GLOBAL_LOGGER_NAME);
+
+    /**
+     * Setup test concerning preset figures in context.
+     *
+     * @param application Application context.
+     */
+    private static void setupPresetTests(Application application) {
+        SelectTestFigureJoeOptionListener selectFirstTestFigureOptionListener = new SelectTestFigureJoeOptionListener(
+            DriverFeature.getDriverManager(), TestOption.FIGURE_JOE1);
```
kowallus, line 35: Confusing and unnecessary format changes (whitespaces)!

src/main/java/edu/kis/powp/command/ComplexCommandBuilder.java
```diff
@@ -0,0 +1,24 @@
...
+        commands = new ArrayList<>();
+    }
+
+    public void addCommand(DriverCommand command){
+        commands.add(command);
+    }
+
+
+    public ComplexCommand build(){
+        return new ComplexCommand(commands);
```
kowallus, line 20: It would be safer to create a copy of commands (so the builder cannot modify it after build)

kowallus (commented):
5 pts OK

PR #139, opened 2020-04-28
First lab

Chester193:


src/main/java/edu/kis/powp/jobs2d/factory/ShapeFactory.java
```diff
@@ -0,0 +1,48 @@
...
+    public static DriverCommand square(Job2dDriver driver) {
+        ComplexCommand command = new ComplexCommand();
+
+        command.addCommand(new SetPositionCommand(0, 0, driver));
+        command.addCommand(new OperateToCommand(0, 60, driver));
+        command.addCommand(new OperateToCommand(60, 60, driver));
+        command.addCommand(new OperateToCommand(60, 0, driver));
+        command.addCommand(new OperateToCommand(0, 0, driver));
+
+        command.execute();
```
kowallus, line 20: factories should return commands without executing them.

src/test/java/edu/kis/powp/jobs2d/TestJobs2dPatterns.java
```diff
@@ -28,8 +30,11 @@ private static void setupPresetTests(Application application) {
 		SelectTestFigureOptionListener selectTestFigureOptionListener = new SelectTestFigureOptionListener(
 				DriverFeature.getDriverManager());
 
-		application.addTest("Figure Joe 1", selectTestFigureOptionListener);
-		application.addTest("Figure Joe 2", selectTestFigureOptionListener);
+		application.addTest(Shape.FIGURE_JOE_1.label, selectTestFigureOptionListener);
+		application.addTest(Shape.FIGURE_JOE_2.label, selectTestFigureOptionListener);
+		application.addTest(Shape.SQUARE.label, event -> ShapeFactory.square(DriverFeature.getDriverManager().getCurrentDriver()));
```
kowallus, line 35: for tests: commands should be executed here.

kowallus (commented):
4 pts 

PR #142, opened 2020-04-29
Command finished

Arrlen:


src/main/java/edu/kis/powp/jobs2d/drivers/command/OperateToCommand.java
```diff
@@ -0,0 +1,22 @@
+package edu.kis.powp.jobs2d.drivers.command;
+import edu.kis.powp.jobs2d.Job2dDriver;
+
+public class OperateToCommand implements DriverCommand{
+    private int x;
+    private int y;
+    private Job2dDriver Driver;
```
kowallus, line 7: in Java convention fields have names starting from a lowercase (driver)

kowallus (commented):
4 pts (too late -1 pts)

PR #143, opened 2020-04-29
Lab 3 Command

Radko-A:


src/main/java/edu/kis/powp/jobs2d/drivers/command/CommandFactory.java
```diff
@@ -0,0 +1,41 @@
+package edu.kis.powp.jobs2d.drivers.command;
+
+import edu.kis.powp.jobs2d.Job2dDriver;
+
+public class CommandFactory {
+
+    public static ComplexCommand addSquare(Job2dDriver driver) {
```
kowallus, line 7: name getSquare would be better

src/main/java/edu/kis/powp/jobs2d/drivers/command/OperateToCommand.java
```diff
@@ -0,0 +1,20 @@
+package edu.kis.powp.jobs2d.drivers.command;
+
+import edu.kis.powp.jobs2d.Job2dDriver;
+
+public class OperateToCommand implements DriverCommand {
+    private int x;
+    private int y;
+    Job2dDriver driver;
```
kowallus, line 8: should be private

kowallus (commented):
3.5 pts (too late -1 pts)

PR #145, opened 2020-05-01
Command 3.1.1 -3.1.5

medykization:


src/test/java/edu/kis/powp/jobs2d/events/SelectTestFigureOptionListener.java
```diff
@@ -11,24 +14,33 @@
...
+			case FIGURE_JOE_2:
+				FiguresJoe.figureScript2(driver);
+				break;
+			case  FIGURE_JANE_1:
+				FiguresJane.figureScript(new DriverAdapter(driver));
 				break;
-			case "Figure Joe 2":
-				FiguresJoe.figureScript2(driverManager.getCurrentDriver());
+			case RECTANGLE:
+				ComplexCommand rectangle = FiguresFactory.getRectangle();
```
kowallus, line 38: Type DriverCommand would be more appropriate here.

kowallus (commented):
4 pts (too late -1 pts)

PR #146, opened 2020-05-03
Lab2 command factory

TheDamik34:


kowallus:
3 pts (too late -2 pts)

PR #147, opened 2020-05-04
Command

michaaeu:


src/main/java/edu/kis/powp/command/factories/TriangleComplexCommandFactory.java
```diff
@@ -0,0 +1,24 @@
...
+import edu.kis.powp.command.OperateToCommand;
+import edu.kis.powp.command.SetPositionCommand;
+import edu.kis.powp.jobs2d.Job2dDriver;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class TriangleComplexCommandFactory implements ComplexCommandFactory {
+    @Override
+    public ComplexCommand getComplexCommand(Job2dDriver job2dDriver) {
```
kowallus, line 14: Returning DriverCommand type would be sufficient (it is usually not important that it is a ComplexCommand)

kowallus (commented):
3 pts (too late)

PR #148, opened 2020-05-06
Command

RadWex:


src/test/java/edu/kis/powp/jobs2d/events/SelectTestFigureOptionListener.java
```diff
@@ -18,8 +20,13 @@ public SelectTestFigureOptionListener(DriverManager driverManager) {
 	public void actionPerformed(ActionEvent e) {
 		if(e.getActionCommand().equals("Figure Joe 1")) {
 			FiguresJoe.figureScript1(driverManager.getCurrentDriver());
-		}else
-		{
+		}else if(e.getActionCommand().equals("Square")){
```
kowallus, line 23: Menu action should not depend on its caption. Enum or separate listeners would be better. At least you should consider constants.

kowallus (commented):
2.5 pts (too late)

PR #149, opened 2020-06-15
215810 Command

tmajcher:
Odpowiedź na pytanie 4 to Kompozyt


src/main/java/edu/kis/powp/jobs2d/drivers/command/ComplexCommand.java
```diff
@@ -0,0 +1,24 @@
+package edu.kis.powp.jobs2d.drivers.command;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class ComplexCommand implements DriverCommand{
+
+    ComplexCommand(){
```
kowallus, line 8: should be public

src/test/java/edu/kis/powp/jobs2d/events/SelectTestFigureOptionListener.java
```diff
@@ -17,5 +17,15 @@ public SelectTestFigureOptionListener(DriverManager driverManager) {
 	@Override
 	public void actionPerformed(ActionEvent e) {
 		FiguresJoe.figureScript1(driverManager.getCurrentDriver());
+		switch(e.getActionCommand()){
+			case "Figure Joe 1":
```

kowallus (commented):
1 pts (min no of points)

